PR #1, opened 2023-11-08
Server setup

AdrianJendo:
Implement template for API and get server started. Default to port 5001 for now.


backend/go.sum
```diff
@@ -0,0 +1,2 @@
+github.com/go-chi/chi/v5 v5.0.10 h1:rLz5avzKpjqxrYwXNfmjkrYYXOyLJd37pz53UFHC6vk=
```
AdrianJendo, line 1: @cameronkinsella let me know if I need to ignore any files like this one.

.idea/lynx-web.iml
```diff
@@ -1,9 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
cameronkinsella: Maybe fixup this commit into a89d68f2c079bd6d75d75778e7bea0d72c7bf608 so that the add and delete of these files won't be shown in the history? It's unrelated to the code written, so I don't think it is valuable to include.

cameronkinsella (changes requested):
Can you add a Go template to the gitignore? It includes common Go artifacts that should be ignored. You can generate one using the ".ignore" plugin in GoLand.

.idea/lynx-web.iml
AdrianJendo: I don't know how to do that so I just force push. Lmk if the gitignore template looks fine.

cameronkinsella (approved):

PR #2, opened 2024-01-17
Aj/frontend setup

AdrianJendo:
Setup frontend boilerplate code


frontend/public/vercel.svg
ABarroso647, line 1: why do we have a vercel image if they aren't sponsoring us? Also, can we add some sort of folder structure for images here

ABarroso647 (approved):
LGTM :shipit:

PR #6, opened 2024-01-29
Add in backend workflow with basic db setup

ABarroso647:
Added features as follows:
Open-api specs to be verified with : https://editor.swagger.io/
Localhost docker setup
User table schema, and initial endpoint,
File structuring for further tables and api endpoints

### DIFF
--- a/backend/config/db.go
+++ b/backend/config/db.go
@@ -9,7 +9,7 @@ import (
 var DB *gorm.DB
 
 func Connect() {
-	db, err := gorm.Open(sqlite.Open("data/database.db"), &gorm.Config{})
+	db, err := gorm.Open(sqlite.Open("data/lynx-web.db"), &gorm.Config{})
 	if err != nil {
 		panic("failed to connect database")
 	}
--- a/backend/config/db.go
+++ b/backend/config/db.go
@@ -1,0 +1,20 @@
+package config
+
+import (
+	"api/models"
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
+var DB *gorm.DB
+
+func Connect() {
+	db, err := gorm.Open(sqlite.Open("data/database.db"), &gorm.Config{})
+	if err != nil {
+		panic("failed to connect database")
+	}
+	// Migrate the schema
+	db.AutoMigrate(&models.User{})
+	DB = db
+
+}
--- a/backend/controller/user.go
+++ b/backend/controller/user.go
@@ -8,7 +8,6 @@ import (
 )
 
 func GetAllUsers(w http.ResponseWriter, r *http.Request) {
-	w.WriteHeader(200)
 	users := []models.User{}
 	config.DB.Find(&users)
 	w.Header().Set("Content-Type", "application/json")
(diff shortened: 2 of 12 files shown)


backend/config/db.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"api/models"
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
+var DB *gorm.DB
+
+func Connect() {
+	db, err := gorm.Open(sqlite.Open("data/database.db"), &gorm.Config{})
```
AdrianJendo: Can we name the db lynx-web or something more meaningful?

backend/controller/user.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"api/config"
+	"api/models"
+	"encoding/json"
+	"net/http"
+)
+
+func GetAllUsers(w http.ResponseWriter, r *http.Request) {
+	w.WriteHeader(200)
```
AdrianJendo: Is this pattern good? Will we still be able to send other status codes for errors (ex: if we can't connect to db)

backend/models/user.go
```diff
@@ -0,0 +1,11 @@
+package models
+
+import "gorm.io/gorm"
+
+type User struct {
+	gorm.Model
+	Id      int    `json:"id" gorm:"primary_key"`
+	Name    string `json:"name"`
+	Email   string `json:"email"`
+	IsAdmin bool   `json:"is_Admin"`
```
AdrianJendo: is_admin

backend/openapi.yml
```diff
@@ -0,0 +1,47 @@
+openapi: 3.0.3
```
AdrianJendo, line 1: I've never used this before, is it just documentation?

backend/server.go
```diff
@@ -13,14 +15,14 @@ const PORT = 5001 // todo: put into environment var
 
 func main() {
 	r := chi.NewRouter()
+	config.Connect()
 	r.Mount("/api", api())
-
+	r.Mount("/api/users", routes.UsersRoute())
```
AdrianJendo: Can we add this to the func below (`func api() chi.Router)`? So that all our API routes are already prefixed with `/api` instead of needing to specify it each time.

AdrianJendo (commented):
Overall looks good

backend/openapi.yml
ABarroso647: Yea if you've ever used like a swagger docs it seems similar. This was a cammy request so essentially before I pr actual code changes that include new endpoints I'll pr the "skeleton" which'll include the location input and output that way there should be less communication issues with what endpoints im making

backend/controller/user.go
ABarroso647: This isn't my end goal for error handling, want to do that mainly all at once with perhaps a middleware I will say if we can't connect to the db the service should panic as per config.Connect()

ABarroso647: image

Will be something like that, which will better handle if we have no response and shit, just want to do all middleware at once, won't expand api functionality without middleware though

AdrianJendo (approved):

PR #10, opened 2024-02-07
Add frontend docker code

AdrianJendo:
Running commands:
```
docker build -t lynx-frontend:dev --target=dev .
docker build -t lynx-frontend:prod --target=prod .
docker run -p 3000:3000 lynx-frontend:<tag>
```

### DIFF
--- a/backend/config/db.go
+++ b/backend/config/db.go
@@ -4,11 +4,18 @@ import (
 	"api/models"
 	"gorm.io/driver/sqlite"
 	"gorm.io/gorm"
+	"os"
+	"path/filepath"
 )
 
 var DB *gorm.DB
 
 func Connect() {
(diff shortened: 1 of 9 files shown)


cameronkinsella:
Made some modifications so the frontend will generate static builds and the backend will serve this static site in production.

AdrianJendo:
@ABarroso647 approve

backend/config/db.go
ABarroso647, line 1: Im confused u seem to be redefining err here???

cameronkinsella: yeah, I'm just not checking for errors on that err. However, since I'm not using it I should have made it `_`. I will fix.

ABarroso647 (approved):

PR #12, opened 2024-02-11
Add in user route functionality

ABarroso647:
Add in all functionality to create, update, and delete users (on their own)

### DIFF
--- a/backend/controller/user.go
+++ b/backend/controller/user.go
@@ -8,7 +8,6 @@ import (
 	"github.com/go-chi/chi/v5"
 	"log"
 	"net/http"
-	"strconv"
 )
 
 func GetAllUsers(w http.ResponseWriter, r *http.Request) {
@@ -63,7 +62,7 @@ func CreateUser(w http.ResponseWriter, r *http.Request) {
 		helpers.DBErrorHandling(result.Error, w, r)
 		return
 	}
-	errJson := json.NewEncoder(w).Encode(r.Host + "/api/users/{" + strconv.Itoa(int(dbUser.ID)) + "}")
+	errJson := json.NewEncoder(w).Encode(dbUser)
 	if errJson != nil {
 		http.Error(w, "500", http.StatusInternalServerError)
 		return
--- a/backend/controller/user.go
+++ b/backend/controller/user.go
@@ -2,14 +2,87 @@ package controller
 
 import (
 	"api/config"
+	"api/helpers"
 	"api/models"
 	"encoding/json"
+	"github.com/go-chi/chi/v5"
+	"log"
 	"net/http"
+	"strconv"
 )
 
 func GetAllUsers(w http.ResponseWriter, r *http.Request) {
-	users := []models.User{}
-	config.DB.Find(&users)
 	w.Header().Set("Content-Type", "application/json")
-	json.NewEncoder(w).Encode(&users)
+	users := []models.User{}
+	result := config.DB.Find(&users)
+	if result.Error != nil {
+		helpers.DBErrorHandling(result.Error, w, r)
+		return
+	}
+	//go yells at you if you don't error check this, better to be safe
+	errJson := json.NewEncoder(w).Encode(&users)
+	if errJson != nil {
+		http.Error(w, "500", http.StatusInternalServerError)
+		return
+	}
+}
+func GetUser(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	uId := chi.URLParam(r, "userId")
+	var user models.User
+	result := config.DB.First(&user, uId)
+
+	if result.Error != nil {
+		helpers.DBErrorHandling(result.Error, w, r)
+		return
+	}
+	errJson := json.NewEncoder(w).Encode(&user)
+	if errJson != nil {
+		http.Error(w, "500", http.StatusInternalServerError)
+		return
+	}
+}
+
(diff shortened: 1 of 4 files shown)


backend/controller/user.go
```diff
@@ -2,14 +2,87 @@ package controller
...
+	if result.Error != nil {
+		helpers.DBErrorHandling(result.Error, w, r)
+		return
+	}
+	//go yells at you if you don't error check this, better to be safe
+	errJson := json.NewEncoder(w).Encode(&users)
+	if errJson != nil {
+		http.Error(w, "500", http.StatusInternalServerError)
+		return
+	}
```
AdrianJendo, line 26: how is data returned?

backend/controller/user.go
```diff
@@ -2,14 +2,87 @@ package controller
...
+	errJson := json.NewEncoder(w).Encode(&user)
+	if errJson != nil {
+		http.Error(w, "500", http.StatusInternalServerError)
+		return
+	}
+}
+
+func CreateUser(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	var user models.UserReq
```
AdrianJendo, line 47: What is UserReq vs. User?

backend/controller/user.go
```diff
@@ -2,14 +2,87 @@ package controller
...
+		if err != nil {
+			log.Print(err)
+		} else {
+			log.Print("Request doesn't contain all params")
+		}
+		http.Error(w, "400", http.StatusBadRequest)
+		return
+	}
+	dbUser := models.User{Name: user.Name,
+		Email: user.Email, IsAdmin: user.IsAdmin}
```
AdrianJendo, line 59: In the future can we do this format:
```
models.User {
    Name: user.Name,
    Email: user.Email,
    IsAdmin: user.IsAdmin,
}
```

backend/controller/user.go
```diff
@@ -2,14 +2,87 @@ package controller
...
+	result := config.DB.Create(&dbUser)
+	if result.Error != nil {
+		helpers.DBErrorHandling(result.Error, w, r)
+		return
+	}
+	errJson := json.NewEncoder(w).Encode(r.Host + "/api/users/{" + strconv.Itoa(int(dbUser.ID)) + "}")
```
AdrianJendo: What does this do?

backend/models/user.go
```diff
@@ -4,8 +4,13 @@ import "gorm.io/gorm"
...
-	IsAdmin bool   `json:"is_admin"`
+	Name    string
+	Email   string
+	IsAdmin bool `json:"default;False"`
+}
+
+type UserReq struct {
+	Name    string `json:"name,omitempty"`
+	Email   string `json:"email,omitempty"`
+	IsAdmin bool   `json:"is_admin,string,default;False"`
```
AdrianJendo, line 15: Is it not possible to combine these into one `User` struct? Why do we have 2.

backend/controller/user.go
ABarroso647: Its in models, essentially userReq contains a list of all the parameters that can be inputted to, (name, email, IsAdmin) and not the non-inputtable params, this way during user input we can verify against the proper params

backend/models/user.go
ABarroso647: See other comment

backend/controller/user.go
ABarroso647: returns a get request link with the newly created userId that can be hit to get the user object, can add a comment for it

backend/controller/user.go
ABarroso647: its not in these lines, but line 21, when we try to write &users, that is what is returning the data

backend/controller/user.go
ABarroso647: Yes

backend/controller/user.go
AdrianJendo: line 23

backend/controller/user.go
AdrianJendo: So we are executing a GET request to get the user object when returning data to the client? We can't just grab the user object directly here?

backend/controller/user.go
AdrianJendo: fine

backend/controller/user.go
ABarroso647: we could if thats the wanted functionality

ABarroso647: depends on the intended use case, I can update it as you think you need

AdrianJendo: This just seems unnecessary steps to me, ill let cameron decide what to do

ABarroso647: he said he'll look into it

cameronkinsella: Apparently the best practice for this is to return the newly created/updated object in the same format you would for a GET request for that object.

AdrianJendo: But can we achieve this without kicking off another request? I've never seen this pattern of actually calling a GET? Like wouldn't even just calling `GetAllUsers` be better?

cameronkinsella: To clarify, I am not talking about kicking off another request, I'm talking about avoiding one.
For example:
POST `/user` creates user with ID 123. Creating this user (the response of the POST request) would return the same response as GET `/user/123`, so you would not need to call  GET `/user/123` after creating it to get its information.

AdrianJendo (approved):
That's much better lol

PR #15, opened 2024-02-16
Being webauthn flow for registering users

AdrianJendo:
Basic webauthn register workflow.

Still needs to implement all the DB connections and saving of data.

https://github.com/lynx-locks/lynx-web/assets/55325093/5c439521-01a3-49f7-9371-b0504d0872b2


backend/controller/auth.go
```diff
@@ -0,0 +1,76 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	var registerReq models.RegisterReq
+	err := json.NewDecoder(r.Body).Decode(&registerReq)
+	if err != nil {
+		log.Print(err)
+		http.Error(w, "400", http.StatusBadRequest)
+		return
+	}
+
+	// TODO: add to DB
```
AdrianJendo: @ABarroso647 help with this part would be appreciated, not familiar with how the db is set up.

backend/controller/auth.go
```diff
@@ -0,0 +1,76 @@
...
+
+	errJson := json.NewEncoder(w).Encode(registerReq) // TODO: send meaningful response
```
AdrianJendo: This part will get updated once we have the key updated in the DB

backend/server.go
```diff
@@ -18,6 +19,11 @@ const PORT = 5001 // todo: put into environment var
 func main() {
 	r := chi.NewRouter()
 	r.Use(middleware.Logger)
+	r.Use(cors.Handler(cors.Options{
+		AllowedOrigins: []string{"http://localhost:3000*"},
```
AdrianJendo, line 23: May be suboptimal but needed CORS errors to go away. Please let me know if I should implement this another way for now.

frontend/src/app/register/page.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export default function RegisterUser() {
+  const searchParams = useSearchParams();
+
+  useEffect(() => {
+    async function f() {
+      // Decode token & check validity
+      const token = searchParams.get("token");
+
+      // TODO: verify token is valid in backend during registerRequest
```
AdrianJendo, line 27: Will be able to implement once we have emails / available_tokens table in backend. I think the verification logic should go in registerRequest.

frontend/src/app/register/page.tsx
```diff
@@ -0,0 +1,88 @@
...
+        // // Obtain transports.
+        // const transports = cred.response.getTransports
+        //   ? cred.response.getTransports()
+        //   : [];
+
+        // credential.response = {
+        //   clientDataJSON,
+        //   attestationObject,
+        //   transports,
+        // };
```
AdrianJendo: I'll need to experiment with these fields once we have tokens & database connection implemented. Not really sure what they are useful for but I will get rid of them once I confirm their use cases.

frontend/src/axios/client.ts
```diff
@@ -0,0 +1,8 @@
+import axios from "axios";
+
+const client = axios.create({
+    baseURL: process.env.API_BASE_URL || "http://localhost:5001"
```
AdrianJendo: Not sure where to add this. Here in dockerfile? I usually use .env files but I don't know if we want to bother setting those up.
Screenshot 2024-02-16 at 12 52 39 AM

ABarroso647: pretty sure @cameronkinsella said he has the env variables already setup, shouldnt need to add anything

backend/routes/auth.go
ABarroso647, line 1: getkeys should just be "/keys" the get seems redundant

backend/models/webAuthn.go
ABarroso647, line 1: are all these models non db related? if they aren't I think this file should probably just be auth.go to match what we have elsewhere. ~~Also, elsewhere we have all params just following the uppercase, so i think stuff like `json:"id"` is probably unnecessary here.~~ Nvm, actually I'll pr with a new db model that replaces gorm.model, that way we can use json returns everywhere

backend/controller/auth.go
ABarroso647: depends what you mean, if you just mean adding to a table to db, you just need to add it to the db you just need to add it to the auto migrate found in db.go example: 
db.Automigrate(&models.User{},
&models.keys{},)

backend/controller/auth.go
```diff
@@ -0,0 +1,76 @@
...
+	if errJson != nil {
+		http.Error(w, "500", http.StatusInternalServerError)
+		return
+	}
+}
+
+func RegisterResponse(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	var registerReq models.RegisterReq
+	err := json.NewDecoder(r.Body).Decode(&registerReq)
```
ABarroso647: in addition, look at the post request for users on how to handle grabbing body variables, afaik what you have here won't error if you are missing some params on your input

backend/server.go
ABarroso647: probably fine for this commit, as long as it doesn't cause any errors with using the other endpoints (specifically can I like still call the endpoints using postman), at the very least we should pull the url from a env variable, not 100% sure on this overall though

frontend/src/axios/client.ts
cameronkinsella: No need to add it anywhere, we just need to document it as a runtime option.
By default prod will be using localhost:5001, so users would only need to change this if they have some kind of advanced configuration.

backend/server.go
cameronkinsella: Can you also confirm that this works without cors errors in prod environment?

backend/controller/auth.go
```diff
@@ -0,0 +1,76 @@
...
+
+	options := models.WebAuthnOptions{
+		Challenge: challenge,
+		Rp: models.Rp{
+			Name: "Lynx Locks",
+		},
+		User: models.UserInfo{
+			Id:          uuid.NewString(), // TODO: get uuid token from url and fetch user id from activeTokens table
+			Name:        "john_james",
+			DisplayName: "John",
```
cameronkinsella: ~~What's the plan for how these values will be set to the real username?~~
Disregard. We should be using the [Go webauthn library](https://github.com/go-webauthn/webauthn) instead of handling this manually

frontend/src/app/register/page.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const searchParams = useSearchParams();
+
+  useEffect(() => {
+    async function f() {
+      // Decode token & check validity
+      const token = searchParams.get("token");
+
+      // TODO: verify token is valid in backend during registerRequest
+
+      const resp = await axios.get(`/api/auth/registerRequest`);
```
cameronkinsella: I know this is what the example used, but could we change the names of these routes?
`/api/auth/registerRequest` -> `/api/auth/register`
`/api/auth/registerResponse` -> `/api/auth/register/verify` or `/api/auth/register/finish`
and then later, for authentication:
`/api/auth/authenticate`
`/api/auth/authenticate/verify` or `/api/auth/authenticate/finish`

backend/models/webAuthnOptions.go
ABarroso647, line 1: also Im wondering the specifics on this "userInfo" is this required to be a separate table or can we use our pre-existing "user" for this

cameronkinsella:
We should consider using a library for the frontend to handle the logic so we can avoid bugs in the future.
For example: https://www.npmjs.com/package/@simplewebauthn/browser

frontend/src/axios/client.ts
AdrianJendo: I see. But either way i feel like leaving it as an environment variable is better?

cameronkinsella: yeah, I agree. I think they way you have it set up is good.
I was just referring to how that env variable should be set. Usually customizable runtime stuff like that would be set by the user instead of in the dockerfile (since the user never actually sees the dockerfile).

From the user's perspective, they would run `docker pull lynx:frontend` and then `docker run -e API_BASE_URL=https://customdomain.com/api lynx:frontend` or something like that.

TLDR: I think the way you did it here is good, we should just document user customizable env variables somewhere

frontend/src/app/register/page.tsx
AdrianJendo: no plz

backend/models/webAuthnOptions.go
AdrianJendo: Not stored in db

AdrianJendo:
> We should consider using a library for the frontend to handle the logic so we can avoid bugs in the future. For example: https://www.npmjs.com/package/@simplewebauthn/browser

I'd rather not use a frontend library if that's ok

frontend/src/app/register/page.tsx
cameronkinsella: Another small change, since the endpoint is customizable we should make the base url in [client.ts](https://github.com/lynx-locks/lynx-web/pull/15/files/277d3ac54a6a27a72fabdce2a0fd794c8ce7807a#diff-1c7b0bd5b820db0dc167ae2b5b90063641dafd1e5ff87b2be4d73c01f9944271) `http://localhost:5001/api/auth` and leave these endpoints as simply `/register`, `/register/verify`, etc.
This way, if someone wants to use a custom domain/routes for their webauthn, they aren't forced to use the full `/api/auth/` base. 

For example, they might want it to be something custom like `https://example.com/api/v1/lynx-auth/register` and `https://example.com/api/v1/lynx-auth/register/verify`.

AdrianJendo: sg

backend/routes/auth.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"api/controller"
+
+	"github.com/go-chi/chi/v5"
+)
+
+func AuthRoute() chi.Router {
+	r := chi.NewRouter()
+	r.Get("/registerRequest", controller.RegisterRequest)
```
cameronkinsella: This should be POST, not GET.
Right now it is using hard-coded info from the backend, but later it will need to identify who is registering, which will require the client to send data.

backend/routes/auth.go
ABarroso647: followup on this file:
looking at other examples like below:
image
Should these not be two post requests, if so, I like the tree logic of them both referring to enroll/register and one being start other being end (or something similar). As itll make better structuring for when you add authenticate and shit... beautiful tree

backend/routes/auth.go
ABarroso647: stop stealing my comments

cameronkinsella: :shipit:

AdrianJendo: I think get is fine, it will send the token as a url param which is all the data we need. Otherwise, I would prefer not to POST since we're not doing any creation/deletion action. Let me know your thoughts.

AdrianJendo: I will change it to POST

backend/controller/auth.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	session := webauthn2.SessionData{
+		Challenge:            sessionData.Challenge,
+		UserID:               sessionData.UserId,
+		AllowedCredentialIDs: sessionData.AllowedCredentialIds,
+		Expires:              sessionData.Expires,
+		UserVerification:     sessionData.UserVerification,
+		Extensions:           sessionData.Extensions,
+	}
```
AdrianJendo, line 109: This library needs a `[]byte` so I wasn't sure if converting our int ids to byte arrays would be sufficient.

backend/models/generic.go
```diff
@@ -1,13 +1,11 @@
...
 import (
-	"gorm.io/gorm"
 	"time"
 )
 
 type Common struct {
-	Id        uint           `gorm:"primaryKey" json:"id"`
-	CreatedAt time.Time      `json:"created_at"`
-	UpdatedAt time.Time      `json:"updated_at"`
-	DeletedAt gorm.DeletedAt `gorm:"index" json:"deleted_at"`
```
AdrianJendo: I left a previous comment about this, what's the point of it

backend/models/user.go
```diff
@@ -1,14 +1,55 @@
...
 
 type UserReq struct {
 	Name    string `json:"name,omitempty"`
 	Email   string `json:"email,omitempty"`
-	IsAdmin bool   `json:"is_admin,string,default;False"`
+	IsAdmin bool   `json:"is_admin,default;False"`
+}
+
+func (u *User) BeforeCreate(tx *gorm.DB) (err error) {
+	webauthnId, err := uuid.NewUUID()
```
AdrianJendo, line 25: I just generate a uuid for each user to use for webauthn because it requires a byte array,

backend/models/generic.go
ABarroso647: Replaces the generic gorm.model that is the whole point...

backend/models/user.go
```diff
@@ -1,14 +1,55 @@
...
+)
+
 type User struct {
 	Common
-	Name    string `json:"name"`
-	Email   string `json:"email"`
-	IsAdmin bool   `json:"is_admin,default;False"`
+	Name       string    `json:"name"`
+	Email      string    `gorm:"unique" json:"email"`
+	IsAdmin    bool      `json:"is_admin,default;False"`
```
cameronkinsella, line 14: ~~Can we change these JSON tags to camelCase in this PR?~~
Disregard, just noticed that you did it in #19.

backend/models/generic.go
AdrianJendo: Does it need the `DeletedAt` field though, this part confuses me because we don't use it.

cameronkinsella: Including the `DeletedAt` field with type `gorm.DeletedAt` changes the behavior of `db.Delete(&object)` to perform a soft delete on that object. Instead of removing the record from the database, gorm will set the `DeletedAt`'s value to the current time and that record will no longer be findable with normal query methods.

I think this feature is normally used for services that have a "recovery" option, like deactivating a social media account. It doesn't seem like we need this functionality for any part of our service, so I think it would be safe to delete just the `DeletedAt` field. What do you think @ABarroso647 ?

backend/webauthn/webauthn.go
ABarroso647, line 1: as this file currently stands, throw this in config

backend/models/webAuthn.go
ABarroso647, line 1: small ting, either fix capitalization if you are planning on adding a different "auth.go" otherwise just make this auth.go imo

backend/config/db.go
ABarroso647, line 1: i think in this case, we should just panic if migrate fails tbh cuz application cant run without it

cameronkinsella:
@AdrianJendo validate frontend check is currently failing because the new changes weren't formatted with prettier.
To automatically fix, just run `npm run fix`.
You can confirm that the status check passes locally with `npm run validate` if needed.

ABarroso647 (approved):

PR #18, opened 2024-02-16
Fix error causing post req to fail on false is_admin creates

ABarroso647:


AdrianJendo (approved):

backend/models/generic.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 type Common struct {
-	ID        uint           `gorm:"primaryKey" json:"id"`
+	Id        uint           `gorm:"primaryKey" json:"id"`
```
AdrianJendo, line 9: After reading https://gorm.io/docs/conventions.html, I understand why you left it as ID. I'm ok using `ID` instead of `Id` if gorm only recognizes `ID` as the default primary key. I believe this will make implementation of foreign keys easier.

AdrianJendo: In which case we could delete `gorm:"primaryKey"`

backend/models/generic.go
```diff
@@ -6,7 +6,7 @@ import (
...
 	CreatedAt time.Time      `json:"created_at"`
 	UpdatedAt time.Time      `json:"updated_at"`
 	DeletedAt gorm.DeletedAt `gorm:"index" json:"deleted_at"`
```
AdrianJendo, line 12: What's use-case of this column? I understand how CreatedAt and UpdatedAt could be useful for most types but when we delete something we're just going to delete that row aren't we?

AdrianJendo (changes requested):

backend/models/generic.go
AdrianJendo: Nevermind, it doesn't care about cases, i made the other change in https://github.com/lynx-locks/lynx-web/pull/19

backend/models/generic.go
AdrianJendo: I removed it in https://github.com/lynx-locks/lynx-web/pull/19, but I can add it back if there is value.

AdrianJendo (approved):

PR #19, opened 2024-02-17
Add all tables

AdrianJendo:
Screenshot 2024-02-17 at 4 03 05 PM

### DIFF
--- a/backend/models/generic.go
+++ b/backend/models/generic.go
@@ -1,13 +1,11 @@
 package models
 
 import (
-	"gorm.io/gorm"
 	"time"
 )
 
 type Common struct {
-	Id        uint           `gorm:"primaryKey" json:"id"`
-	CreatedAt time.Time      `json:"created_at"`
-	UpdatedAt time.Time      `json:"updated_at"`
-	DeletedAt gorm.DeletedAt `gorm:"index" json:"deleted_at"`
+	Id        uint      `json:"id"`
+	CreatedAt time.Time `json:"createdAt"`
+	UpdatedAt time.Time `json:"updatedAt"`
 }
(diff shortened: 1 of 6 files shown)


backend/models/key.go
ABarroso647, line 1: Why does this both have a userId and a user?

AdrianJendo: I just followed this doc https://gorm.io/docs/belongs_to.html

ABarroso647 (dismissed):
I would like config db to be separated line by line but that can be fixed in future small enuf Ill still approve

backend/models/generic.go
```diff
@@ -1,13 +1,11 @@
...
-	"gorm.io/gorm"
 	"time"
 )
 
 type Common struct {
-	ID        uint           `gorm:"primaryKey" json:"id"`
-	CreatedAt time.Time      `json:"created_at"`
-	UpdatedAt time.Time      `json:"updated_at"`
-	DeletedAt gorm.DeletedAt `gorm:"index" json:"deleted_at"`
+	Id        uint      `json:"id"`
```
ABarroso647, line 8: why is id no longer a primary key?

AdrianJendo: It still is, you don't need to specify `gorm:"primaryKey"` if the field name is Id: https://gorm.io/docs/models.html#Conventions. You can also see in the screenshot I posted for the descrition.

ABarroso647: figured was case sensitive, makes sense

ABarroso647 (dismissed):

cameronkinsella (approved):

cameronkinsella (approved):

PR #22, opened 2024-02-19
 Feature/adding generics

ABarroso647:
Update basic crud calls so they can be done generically for all tables

### DIFF
--- a/backend/controller/user.go
+++ b/backend/controller/user.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"github.com/go-chi/chi/v5"
 	"net/http"
+	"strconv"
 )
 
 func GetAllUsers(w http.ResponseWriter, _ *http.Request) {
@@ -14,8 +15,14 @@ func GetAllUsers(w http.ResponseWriter, _ *http.Request) {
 }
 func GetUser(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
-	uId := chi.URLParam(r, "userId")
-	helpers.GetById(w, models.User{}, uId)
+
+	uId, err := strconv.ParseUint(chi.URLParam(r, "userId"), 10, 32)
+	if err != nil {
+		http.Error(w, "400", http.StatusBadRequest)
+		return
+	}
+	//extra uint is for 64 bit to 32 bit
+	helpers.GetFirstTable(w, models.User{}, models.Common{Id: uint(uId)})
 }
 
 func CreateUser(w http.ResponseWriter, r *http.Request) {
--- a/backend/controller/user.go
+++ b/backend/controller/user.go
@@ -1,87 +1,34 @@
 package controller
 
 import (
-	"api/config"
 	"api/helpers"
 	"api/models"
 	"encoding/json"
 	"github.com/go-chi/chi/v5"
-	"log"
 	"net/http"
 )
 
-func GetAllUsers(w http.ResponseWriter, r *http.Request) {
+func GetAllUsers(w http.ResponseWriter, _ *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
-	users := []models.User{}
-	result := config.DB.Find(&users)
-	if result.Error != nil {
-		helpers.DBErrorHandling(result.Error, w, r)
-		return
-	}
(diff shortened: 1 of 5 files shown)


backend/controller/user.go
```diff
@@ -1,87 +1,33 @@
...
-		helpers.DBErrorHandling(result.Error, w, r)
-		return
-	}
-	errJson := json.NewEncoder(w).Encode(dbUser)
-	if errJson != nil {
-		http.Error(w, "500", http.StatusInternalServerError)
-		return
-	}
-
+	user.Common = models.Common{}
```
ABarroso647, line 25: essentially ignore any inputs the user made to anything within "common", which should not be writeable

backend/controller/user.go
```diff
@@ -1,87 +1,34 @@
...
-		return
-	}
-	errJson := json.NewEncoder(w).Encode(dbUser)
-	if errJson != nil {
-		http.Error(w, "500", http.StatusInternalServerError)
-		return
-	}
-
+	// essentially reset if the user inputted anything to common as it should not be editable
+	user.Common = models.Common{}
```
ABarroso647, line 33: if anyone can think of a cleaner way to do this im open

AdrianJendo: What's the point of this

backend/helpers/helpers.go
```diff
@@ -13,24 +17,91 @@ func CheckErr(err error) {
...
+	}
+	//go yells at you if you don't error check this, better to be safe
+	errJson := json.NewEncoder(w).Encode(&table)
+	if errJson != nil {
+		http.Error(w, "500", http.StatusInternalServerError)
+		return
+	}
+}
+
+func GetById[T models.AllTables](w http.ResponseWriter, table T, Id string) {
```
AdrianJendo: Is this get by Id or get by primaryKey?

backend/models/generic.go
```diff
@@ -5,7 +5,11 @@ import (
 )
 
 type Common struct {
-	Id        uint      `json:"id"`
+	Id        uint      `json:"id" gorm:"->"`
```
AdrianJendo, line 8: What is this for

backend/controller/user.go
ABarroso647: same function as what the user_req object used to do, makes it so people cannot edit "created_at" and "updated_at", can't set these to readonly like id, as then gorm will not automatically update them (kinda stupid).

backend/helpers/helpers.go
ABarroso647: it'd be primary key, can rename, in our case i think all our primary keys are some form of Id based on atleast the common structure so I named it as so

backend/models/generic.go
ABarroso647: this just makes it readonly, technically this is covered by line 26 in backend/controller.go but it is just technically a more proper definition of Id as we should never  be manually updating it, so ive left it as so

backend/helpers/helpers.go
AdrianJendo: lets rename it primaryKey i think that's more obvious

AdrianJendo: on another note why is it a string, aren't out ids ints

ABarroso647: gonna make it pk, just cuz go variable naming. 
It works the practically the same if its a string or an int, I may need to cast and like manually handle the string-> int if I make it an integer (which I'd want to do in this function anyways cuz it'd be the same for every pk so I think in that case I'd probably still want to pass it as a string and then handle it as integer). The db will basically spit out key not found if you pass it a string so that's why I left it as opposed to me manually spitting out "id is not an integer" on error so. Lmk if you care enough.

ABarroso647: Nvm, see new commit for updates to this

backend/models/generic.go
cameronkinsella: Why aren't we using `gorm:"primaryKey"` again?

ABarroso647: i didnt make this change, see adrian's commend b4, basically Id is auto recognized as primary key

cameronkinsella (approved):
:shipit:

PR #24, opened 2024-02-24
Implement webauthn for authenticating users

AdrianJendo:
Here a high quality video of it in action.

https://github.com/lynx-locks/lynx-web/assets/55325093/c945aea9-b4df-4b9e-afeb-b9378d932a54


backend/config/db.go
```diff
@@ -21,9 +20,9 @@ func Connect() {
 		panic("failed to connect database")
 	}
 	// Migrate the schema
-	db.AutoMigrate(&models.User{}, &models.Key{}, &models.Role{}, &models.ActiveTokens{}, &models.Door{}, &models.SessionData{})
+	db.AutoMigrate(&models.User{}, &models.Credential{}, &models.Role{}, &models.ActiveTokens{}, &models.Door{}, &models.SessionData{}, &models.Credential{})
```
AdrianJendo: Renamed key table to credential table because i need to store other credential-related items in a table and naturally it didn't make sense to have a key table and a credential table where the key table stores a subset of the credential table. Can rename back to Key but not sure it makes sense since the lingo gets more confusing.

backend/controller/auth.go
```diff
@@ -59,41 +57,156 @@ func RegisterRequest(w http.ResponseWriter, r *http.Request) {
...
 
-	var completeRegistration models.CompleteRegistration
-	err = json.NewDecoder(bytes.NewReader(contents)).Decode(&completeRegistration)
+	// Get the session data stored from the function above
+	sessionData := models.SessionData{
+		UserId: user.WebAuthnID(),
+	}
+	result = db.DB.First(&sessionData)
+	db.DB.Delete(&sessionData)
+	if result.Error != nil {
```
AdrianJendo, line 92: In a real world this should probably be in a cache or stored some other way but time constrained

backend/controller/auth.go
```diff
@@ -59,41 +57,156 @@ func RegisterRequest(w http.ResponseWriter, r *http.Request) {
...
+		helpers.DBErrorHandling(result.Error, w)
+		return
+	}
 
+	session := webauthn2.SessionData{
```
AdrianJendo, line 97: i needed to implement this because the `SessionData` can't be stored unless it includes the `gorm:"serializer:json"` annotation.

Screenshot 2024-02-23 at 9 21 04 PM

backend/db/db.go
```diff
@@ -0,0 +1,5 @@
+package db
+
+import "gorm.io/gorm"
+
+var DB *gorm.DB
```
AdrianJendo, line 5: We need to access DB in user model so leaving this in config makes a circular dependency.

backend/routes/auth.go
```diff
@@ -8,8 +8,9 @@ import (
 
 func AuthRoute() chi.Router {
 	r := chi.NewRouter()
-	r.Post("/register/request", controller.RegisterRequest)
-	r.Post("/register/response", controller.RegisterResponse)
-	r.Get("/keys", controller.GetKeys)
+	r.Post("/register/request/{userId}", controller.RegisterRequest)
```
AdrianJendo, line 11: I added the user to the route because a) we don't really care if we expose this data and b) the webauthn functions need to read from the POST request body, and because the request body is an IO stream, it's consumed whenever it is read. Ie: it is annoying to read the request body multiple times (once for getting user id, once for getting remaining webauthn data)

backend/controller/auth.go
```diff
@@ -2,32 +2,48 @@ package controller
...
-	"github.com/google/uuid"
-	"io"
-	"log"
 	"net/http"
+	"strconv"
 )
 
 // https://developers.google.com/codelabs/passkey-form-autofill
 
+func getUser(w http.ResponseWriter, r *http.Request) (user models.User, valid bool) {
```
ABarroso647, line 17: this is fine for this pr, but will be replaced in future

backend/db/db.go
ABarroso647: Is this like a webauthn thing, I dont see the circular dependency ngl

AdrianJendo: The user model requires the database, so leaving this in config would result in user model importing config and config importing user model

ABarroso647 (approved):

PR #25, opened 2024-02-25
Add role and door functionality

ABarroso647:


backend/models/door.go
```diff
@@ -2,7 +2,7 @@ package models
 
 type Door struct {
 	Common
-	Name        string `json:"name"`
-	Description string `json:"description"`
-	Roles       []*Key `gorm:"many2many:role_door;"`
+	Name        string  `json:"name"`
+	Description string  `json:"description"`
+	Roles       []*Role `gorm:"many2many:role_door;"`
```
AdrianJendo: oops thx

backend/helpers/helpers.go
```diff
@@ -42,66 +42,62 @@ func DBErrorHandling(error error, w http.ResponseWriter) {
...
-func DeleteById[T models.AllTables](w http.ResponseWriter, table T, Id string) {
+func DeleteById[T models.AllTables](w http.ResponseWriter, table T, Id string) error {
 	result := config.DB.Unscoped().Delete(&table, Id)
 	if result.Error != nil {
 		DBErrorHandling(result.Error, w)
-		return
+		return result.Error
 	}
 	if result.RowsAffected < 1 {
 		http.Error(w, "404", http.StatusBadRequest)
```
AdrianJendo: can we have more meaningful error messages instead of "404", which is already defined in `http.StatusBadRequest`

AdrianJendo: There's a few places where this can be improved but i'm just going to make this one comment.

backend/controller/user.go
```diff
@@ -31,11 +37,17 @@ func CreateUser(w http.ResponseWriter, r *http.Request) {
 	err := json.NewDecoder(r.Body).Decode(&user)
 	// essentially reset if the user inputted anything to common as it should not be editable
 	user.Common = models.Common{}
-	helpers.CreateNewRecord(w, user, err)
+	err2, user := helpers.CreateNewRecord(w, user, err)
```
AdrianJendo: i typically just redefine `err, user =` i don't know if that's less confusing / changes readability in any way but that's what ive been doing.

backend/controller/role.go
```diff
@@ -0,0 +1,94 @@
...
+func DeleteRole(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	rId := chi.URLParam(r, "roleId")
+	err := helpers.DeleteById(w, models.Role{}, rId)
+	if err == nil {
+		w.WriteHeader(200)
+	}
+}
+
+func UpsertDoorAssociation(w http.ResponseWriter, r *http.Request) {
```
AdrianJendo: fancy

AdrianJendo (commented):
Minor things, otherwise lgtm. Can we merge this after the webauthn stuff because you need to update the `config.DB` usages

backend/controller/door.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/go-chi/chi/v5"
+	"net/http"
+)
+
+func GetAllDoors(w http.ResponseWriter, _ *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	err, doors := helpers.GetAllTable(w, []models.Door{})
+	if err == nil {
+		helpers.JsonWriter(w, doors)
+	}
```
AdrianJendo, line 16: Shouldn't we do something if err != nil?

backend/controller/door.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func CreateDoor(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	var role models.Door
+	err := json.NewDecoder(r.Body).Decode(&role)
```
AdrianJendo, line 23: Why would you define err if its unused? Also if the client sends incorrectly formatted payload we should tell them.

backend/controller/door.go
```diff
@@ -0,0 +1,38 @@
...
+	role.Common = models.Common{}
+	err, user := helpers.CreateNewRecord(w, role, err)
+	if err == nil {
+		helpers.JsonWriter(w, user)
+	}
+}
+
+func DeleteDoor(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	dId := chi.URLParam(r, "doorId")
```
AdrianJendo: should we check null or does deletebyid error report a sufficient error if its null?

backend/controller/role.go
```diff
@@ -0,0 +1,98 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	err, roles := helpers.GetAllTable(w, []models.Role{})
+	if err == nil {
+		helpers.JsonWriter(w, roles)
+	}
+
+	//for viewing all connections
+	//roles = []models.Role{}
+	//config.DB.Preload("Doors").Find(&roles)
+	//helpers.JsonWriter(w, roles)
```
AdrianJendo: can we delete this?

backend/controller/role.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func CreateRole(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	var role models.Role
+	err := json.NewDecoder(r.Body).Decode(&role)
+	role.Common = models.Common{}
```
AdrianJendo, line 28: I'm still confused by this, maybe you can explain better here: does instantiating a model not populate this by default?

What values get populated into common? Does it default CreatedAt and UpdatedAt to the current time? Isn't the Id set when the model is created (this is the main reason I'm asking about this), so what functionality does this add?

Also maybe we can write a `BeforeCreate` for our models to specify it there instead of here? If this sounds like a good approach, let's open up a follow up PR for this.

backend/controller/role.go
```diff
@@ -0,0 +1,98 @@
...
+	doors := []models.Door{}
+
+	if len(doorIDs) == 0 {
+		http.Error(w, "400 missing door ids", http.StatusBadRequest)
+		return
+	}
+	results := db.DB.Where(&doorIDs).Find(&doors)
+
+	if results.RowsAffected != int64(len(doorIDs)) {
+		http.Error(w, "40 one or more invalid door id/s", http.StatusBadRequest)
```
AdrianJendo: 400

backend/controller/role.go
```diff
@@ -0,0 +1,98 @@
...
+	if err == nil {
+		w.WriteHeader(200)
+	}
+}
+
+func ReplaceDoorAssociation(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	err, rId := helpers.ParseInt(w, r, "roleId")
+	if err != nil {
+		return
```
AdrianJendo, line 58: do we want to send back a response?

backend/controller/door.go
AdrianJendo: I see you use this pattern in a lot of places. I think writing a function to handle this (and also handle an error) would be a good idea.

backend/models/role.go
```diff
@@ -2,7 +2,6 @@ package models
 
 type Role struct {
 	Common
-	Name string        `json:"name"`
-	Keys []*Credential `gorm:"many2many:key_role;"`
-	Door []*Door       `gorm:"many2many:role_door;"`
+	Name  string `json:"name"`
+	Doors []Door `json:"doors,omitempty" gorm:"many2many:role_door;"`
```
AdrianJendo: What is there not more * (but the Door model still has it). Also why did you get rid of the credentials connection?

backend/controller/door.go
ABarroso647: yes because, the getAlltables function returns if it had an error, if it had an error it will handle it but then obv u dont wna call the writer if there was an error (works the same as return on error). I think a function is redundant in that case for these reasons

backend/controller/door.go
ABarroso647: Its used read the create new record function, I could probably rework the logic but you'll notice I passed it in, and thats cuz I wanted the create new Records function to be able to handle cases where you got the req body other ways and hence removing that logic from it, but this error is handled if you look in there

backend/controller/door.go
ABarroso647: Wdym? its a url param, im 99% sure that throws a 405 if you just dont pass a param, if you like put /null itll say record not found

backend/controller/role.go
ABarroso647: potentially, I found myself using it a lot for validation hence keeping it just much easier to check everything is in order by running a get all

backend/controller/role.go
ABarroso647: its handled in the parseint

backend/controller/role.go
ABarroso647: A before create maybe, but essentially I just dont want users to be able to called the funciton with request body as follows
{
"id": 10000,
"CreatedAt": (some time maybe yesterday),
"Name": John,
...
}

Essentially this just resets those common values to default to ignore any population a user would attempt to do since gorm should be the only thing populating those fields and since we cant set those fields to read only. Also since looping over fields in goLang is like annoying cuz of reflection this avoids having to do any looping. But perhaps this could be done in a before create, could be a clean way didnt think of that

backend/controller/role.go
ABarroso647: :(

backend/models/role.go
ABarroso647: Its unneccessary, the connection is still there if you look at Gorm https://gorm.io/docs/many_to_many.html for many to many we actually only need the definition in one table. The only benefit of having it as a back-reference is if you want to look both ways, but in this case we won't ever be looking from roles to keys just to figure out what roles a key has. Also the * seemed useless to me??? lmk if you had a specific reason, I should remove it from both sides though so its consistent

ABarroso647: For reference I left the Roles connection in doors, even though we aren't getting the role associations with a door cuz I could see us doing that in the future like a lil map with who has access to what door or smthn but even then its probably unneccessary, we really don't need the back-reference tbh

backend/controller/door.go
ABarroso647: This  is the case for most of the helper functions, in general the idea is they'll handle all their own errors and then return if they errored like in basic what the error was, so all this fn need to do is go "oh shit it errored dont continue"

backend/controller/door.go
AdrianJendo: I see

backend/controller/door.go
AdrianJendo: What if it's formatted incorrectly?

ABarroso647: works the exact same way as user does, it'll error and won't delete shit and throw the user an eror, I can just add in the parseInt function if we are concerned, but it will error, it just may not be as specific as if I had that fn in there. But this is the same way its done in user so idk if there's been issues using delete for user currently?

backend/controller/door.go
cameronkinsella: Bit of a nitpick, but it's more idiomatic to handle the error with `err != nil` and `return` early if necessary. Not a big deal though.

backend/models/role.go
ABarroso647: Ended up adding it back in based on recent commit, although we don't need it for gets having it there so we can hard delete associations on deletion is nice (still don't see the point/value in the * here tho)

AdrianJendo (approved):

PR #26, opened 2024-02-29
Implement frontend authorization

AdrianJendo:
Login with local storage to save token.

https://github.com/lynx-locks/lynx-web/assets/55325093/eccc3a5e-0e38-49e3-9ea0-ef6450d58ac1


backend/controller/adminAuth.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"api/models"
+	"encoding/json"
+	"github.com/golang-jwt/jwt/v5"
+	"net/http"
+	"time"
+)
+
+var secretKey = []byte("secret") // TODO: put in .env file
```
AdrianJendo: This should go in a .env file

backend/controller/webAuthn.go
```diff
@@ -32,11 +33,73 @@ func getUser(w http.ResponseWriter, r *http.Request) (user models.User, valid bo
 	return
 }
 
+func ExecAuthResponse(w http.ResponseWriter, r *http.Request, user models.User) error {
```
AdrianJendo, line 36: moved some repeated logic to a function.

AdrianJendo (commented):
It's not the most flawless authentication logic, but I think it's good enough for now given the simplicity of the application. I would consider revisiting this at a later date to clean it up, but at least the frontend routes are now protected.

I don't think it's necessary for the MVP but theoretically we should protect backend routes so that an attacker can't interact with unauthorized endpoints.

Obviously this change is much more important for the frontend because people will be accessing certain pages of this app (such as `/register` and `/authorize`) so we don't want them accidentally navigating to the admin screen.

backend/controller/adminAuth.go
cameronkinsella: Will this break with an empty string? If not, I think it would be worth replacing this with `[]byte(os.Getenv("TOKEN_SECRET"))` in this PR. This will use the environment variable if available, but will return an empty string if not.

ABarroso647: +1

backend/controller/adminAuth.go
ABarroso647, line 1: just call this file "login" or something i think the long name made sense when u had webauthn but like i dont see the value for this

backend/controller/user.go
```diff
@@ -26,6 +26,24 @@ func GetUser(w http.ResponseWriter, r *http.Request) {
 	helpers.JsonWriter(w, user)
 }
 
+func GetUserByEmail(w http.ResponseWriter, r *http.Request) {
```
ABarroso647, line 29: this isn't necessary you can use the getFirstTable function in the helper function as long as you pass the param

backend/routes/user.go
```diff
@@ -9,6 +9,7 @@ func UsersRoute() chi.Router {
 	r := chi.NewRouter()
 	r.Get("/", controller.GetAllUsers)
 	r.Get("/{userId}", controller.GetUser)
+	r.Get("/login", controller.GetUserByEmail)
```
ABarroso647, line 12: this should just be a url param im pretty sure @cameronkinsella  thoughts?

backend/controller/user.go
AdrianJendo: ? not following, i believe this function is necessary because we have no other endpoint to get a user by email

backend/controller/adminAuth.go
AdrianJendo: @cameronkinsella did you fix getting the secret from the frontend?

backend/controller/user.go
ABarroso647: Get first table takes a param, if you look at getby user id or pk or whatever we pass the userId but you can just as easily pass email instead and it functions the same

ABarroso647: I can put sample function when i get home

backend/routes/user.go
cameronkinsella: +1 iirc it's not RESTful for GET to have a body

backend/controller/adminAuth.go
AdrianJendo: Also this doesn't work for an empty string i forgot to check earlier

backend/routes/user.go
AdrianJendo: it's not a body. It's a query parameter:
```http://localhost:5001/api/users/login/?email=...```

File: backend/controller/user.go (hunk 3 above)
Comment (AdrianJendo) [reply backend/controller/user.go:29]: That's what im doing in this function so im confused what you mean

File: backend/routes/user.go (hunk 4 above)
Comment (cameronkinsella) [reply backend/routes/user.go:12]: Oh, my bad. LGTM

File: backend/controller/adminAuth.go (hunk 1 above)
Comment (cameronkinsella) [reply backend/controller/adminAuth.go]: > @cameronkinsella did you fix getting the secret from the frontend?

Not sure what you're referring to. Can you elaborate?

Comment (cameronkinsella) [reply backend/controller/adminAuth.go]: > Also this doesn't work for an empty string i forgot to check earlier

Okay, if it doesn't work with "" instead of "secret" then I think it's fine to leave it as "secret" for now. 

We can add a helper function in a later PR to manage env variables with default values a bit better. We can use it for what I added in #27 to clean that up as well.

File: backend/controller/user.go [hunk 5]
```diff
@@ -26,6 +26,24 @@ func GetUser(w http.ResponseWriter, r *http.Request) {
...
+
+	email := r.URL.Query().Get("email")
+	if email == "" {
+		http.Error(w, "Email parameter is required", http.StatusBadRequest)
+		return
+	}
+
+	err, user := helpers.GetFirstTable(w, models.User{}, models.User{Email: email})
+	if err != nil {
+		helpers.DBErrorHandling(err, w)
```
ABarroso647: @AdrianJendo  sorry I missed you using the first function cuz I saw this error handling. This shoukd be redundant the function already does this so im confused why its being called again

backend/routes/user.go
ABarroso647: Same... I was comparing to the other query params ig this is just handled diff cuz its a email

backend/controller/user.go
ABarroso647: Specifical the dbe error handling call

backend/controller/adminAuth.go
AdrianJendo: > > @cameronkinsella did you fix getting the secret from the frontend?
> 
> Not sure what you're referring to. Can you elaborate?

i thought you mentioned there was an issue getting environment variables into the frontend at one point.

cameronkinsella: Oh, yeah that's right. The fix for frontend environment variables was added in #27. I added a guide for how to add new ones in the frontend's README.

AdrianJendo: ok cool, not changing anything for this PR though

cameronkinsella: yeah, that's fine.

backend/controller/user.go
AdrianJendo: Done

ABarroso647 (approved):

PR #27, opened 2024-03-01
Use TLS for production

cameronkinsella:
TLS is mandatory for the WebAuthn protocol, so production now uses `autocert` to automatically create and renew certs for the given domain.

Also fixed issue where environment variables weren't being updated on runtime for frontend.

https://app.lynx-locks.com/


backend/config/webauthn.go
```diff
@@ -10,11 +10,12 @@ var (
...
 )
 
+// TODO make these values configurable via environment variables
 func init() {
 	wconfig := &webauthn.Config{
-		RPDisplayName: "Lynx Locks",                                               // Display Name for your site
-		RPID:          "localhost",                                                // Generally the FQDN for your site
-		RPOrigins:     []string{"http://localhost:3000", "http://localhost:5001"}, // The origin URLs allowed for WebAuthn requests
+		RPDisplayName: "Lynx Locks",                                                                             // Display Name for your site
+		RPID:          "app.lynx-locks.com",                                                                     // Generally the FQDN for your site
```
AdrianJendo: We should load this in from the environment instead of hardcoding, i still want webauthn to work on localhost

cameronkinsella: I know, this is in the TODO on line 13. I'll just add it to this PR.

AdrianJendo: oh sorry i didnt see the TODO. Yeah let's add it here so that i can still use it locally without any changes

cameronkinsella: Pushed a new commit with these changes.
The domain is now configurable in prod, but will use localhost for dev and app.lynx-locks.com for prod by default if not specified.

AdrianJendo: awesome

backend/config/webauthn.go
```diff
@@ -11,10 +12,25 @@ var (
...
+	if value, ok := os.LookupEnv("NODE_ENV"); ok && value == "production" {
+		if domain, ok := os.LookupEnv("WEBAUTHN_DOMAIN"); ok {
+			id = domain
+			origins = []string{fmt.Sprintf("https://%s", domain)}
+		} else {
+			id = "app.lynx-locks.com"
+			origins = []string{"https://app.lynx-locks.com"}
+		}
+	} else {
+		id = "http://localhost:5001"
```
AdrianJendo: Must be `localhost` without port or http:// i believe

backend/server.go
```diff
@@ -8,29 +8,97 @@ import (
...
+			}
+
+			return nil
+		})
+
+		if err != nil {
+			fmt.Printf("Error walking the path: %v\n", err)
+		}
+	}
+}
```
AdrianJendo: can you just comment on what this is doing

backend/server.go
```diff
@@ -8,29 +8,97 @@ import (
...
 
 func main() {
+	var origins []string
+	if value, ok := os.LookupEnv("NODE_ENV"); ok && value == "production" {
+		if domain, ok := os.LookupEnv("WEBAUTHN_DOMAIN"); ok {
```
AdrianJendo: maybe we can call this `APP_DOMAIN` or `CLIENT_DOMAIN` something, since its going to cors?

backend/server.go
```diff
@@ -8,29 +8,97 @@ import (
...
-	err := http.ListenAndServe(fmt.Sprintf(":%d", PORT), r)
-	helpers.CheckErr(err)
+	var PORT int
+	if value, ok := os.LookupEnv("NODE_ENV"); ok && value == "production" {
+		PORT = 443
+
+		m := &autocert.Manager{
+			Cache:      autocert.DirCache("cert-cache"),
+			Prompt:     autocert.AcceptTOS,
+			HostPolicy: autocert.HostWhitelist("app.lynx-locks.com"),
```
AdrianJendo: Should we try to read this from ENV var just like line 54?

backend/server.go
```diff
@@ -8,29 +8,97 @@ import (
...
 		AllowedHeaders: []string{"Accept", "Authorization", "Content-Type", "X-CSRF-Token"},
 	}))
 	config.Connect()
 	FileServer(r, "/", http.Dir("./static"))
 	r.Mount("/api", api())
 
-	fmt.Printf("Server running on port %d\n", PORT)
-	err := http.ListenAndServe(fmt.Sprintf(":%d", PORT), r)
-	helpers.CheckErr(err)
+	var PORT int
```
AdrianJendo: why are we declaring the PORT up here if its only used in scope?

backend/server.go
```diff
@@ -8,29 +8,97 @@ import (
...
+
+		server := &http.Server{
+			Addr:      fmt.Sprintf(":%d", PORT),
+			Handler:   r,
+			TLSConfig: m.TLSConfig(),
+		}
+
+		fmt.Printf("Server running on port %d\n", PORT)
+		go func() {
+			err := http.ListenAndServe(":80", m.HTTPHandler(nil))
```
AdrianJendo, line 95: Why do we have port 80 and port 443 here?

cameronkinsella: This is for communication for the http-01 challenge for certs. I'll add a comment.

AdrianJendo (approved):
Good stuff, i like the comments

PR #28, opened 2024-03-03
Fixup backend structure, allow user, door, and role updates, and add associations between users and roles

ABarroso647:
Sample request:
image
response:
image

Invalid request cuz u didn't add id
image
response:
image


backend/config/db.go
```diff
@@ -15,7 +15,7 @@ func Connect() {
 	if err != nil {
 		panic("failed to create data directory")
 	}
-	db, err := gorm.Open(sqlite.Open("data/lynx-web.db"), &gorm.Config{})
+	db, err := gorm.Open(sqlite.Open("data/lynx-web.db"), &gorm.Config{TranslateError: true})
```
ABarroso647, line 18: This allows extra error checking such as checking unique key

backend/controller/door.go
```diff
@@ -17,6 +17,25 @@ func GetAllDoors(w http.ResponseWriter, _ *http.Request) {
...
+
+	err, dId := helpers.ParseInt(w, r, "doorId")
+	if err != nil {
+		return
+	}
+	err, door := helpers.GetFirstTable(w, models.Door{}, models.Common{Id: dId})
+	if err != nil {
+		return
+	}
+	err, door = helpers.UpdateSpecifiedParams(w, r, &door, door.Common, &door.Common)
```
ABarroso647: the pointer shit here is like kinda confusing, but essentially im just using it as an agnostic way to update the common variable in the helper function even though we can't explicity call common, to check if the user passed any values to common

backend/helpers/helpers.go
```diff
@@ -72,9 +73,27 @@ func GetFirstTable[T models.AllTables, P models.AllTables](w http.ResponseWriter
...
+		DBErrorHandling(result.Error, w)
+		return result.Error, *table
+	}
+	return nil, *table
+}
+
 func CreateNewRecord[T models.AllTables](w http.ResponseWriter, table T) (error, T) {
 	if !CheckEmptyString(table) {
-		log.Print("Request contains empty string params")
+		http.Error(w, "400 Request contains empty string params", http.StatusBadRequest)
```
ABarroso647: lol

backend/models/user.go
```diff
@@ -13,7 +13,7 @@ type User struct {
 	Common
 	Name       string `json:"name"`
 	Email      string `gorm:"unique" json:"email"`
-	IsAdmin    bool   `json:"isAdmin,default;False"`
+	IsAdmin    bool   `json:"isAdmin"`
```
ABarroso647: ok so i verified this with createUser and you can still not specify it on create and it will default to zero, but it was causing issues where we couldn't know whether or not to update IsAdmin since that bitch could be the default value or the user could've just put false in, so this saved a lot of specific logic for bools and I dont think breaks anything (I didn't verify your auth functions though)

backend/routes/door.go
```diff
@@ -9,6 +9,7 @@ func DoorsRoute() chi.Router {
 	r := chi.NewRouter()
 	r.Get("/", controller.GetAllDoors)
 	r.Post("/", controller.CreateDoor)
+	r.Patch("/{doorId}", controller.UpdateDoor)
```
ABarroso647: patch > put cuz we aren't always updating every value and u dont need to pass every param to do an update, I set the Id as queryparam for all of these cuz its super annoying to parse

backend/controller/door.go
AdrianJendo: I don't really like this, is there another way we can implement it? I think the more we develop this, the more confusing this 'Common' model is becoming. I also noticed we don't use the `embed` tag anywhere, as per gorm docs. Is there a reason we don't do this. Either way, I think refactoring Common to not require these workaround is justifiable, or refactoring this code to not continue to use weird workarounds.

Screenshot 2024-03-03 at 1 46 47 PM

backend/helpers/helpers.go
```diff
@@ -72,9 +73,27 @@ func GetFirstTable[T models.AllTables, P models.AllTables](w http.ResponseWriter
...
+func UpdateSpecifiedParams[T models.AllTables](w http.ResponseWriter, r *http.Request, table *T, initCommon models.Common, updatedCommon *models.Common) (error, T) {
+	err := json.NewDecoder(r.Body).Decode(&table)
+	if err != nil {
+		http.Error(w, "400 malformed request", http.StatusBadRequest)
+		return err, *table
+	}
+	if initCommon != *updatedCommon {
+		http.Error(w, "Invalid updates to automatic fields", http.StatusBadRequest)
+		return errors.New("400"), *table
+	}
```
AdrianJendo: This check is confusing because it's not intuitive why initCommon would ever != updatedCommon

backend/helpers/helpers.go
```diff
@@ -72,9 +73,27 @@ func GetFirstTable[T models.AllTables, P models.AllTables](w http.ResponseWriter
...
+	result := db.DB.Debug().Save(&table)
```
AdrianJendo: Why does this have Debug()

backend/models/user.go
AdrianJendo: I think keeping default False is better for readability. If we override a False with a False i don't see that as a problem. As long as the isAdmin flag doesn't randomly change between updates I think it's fine.

backend/controller/door.go
AdrianJendo: Also the only Common variable that should get updated is the `UpdatedAt` one, can't we just set that some other way. I think we should look into readonly fields to handle not updating the `CreatedAt` (and the Id, but if we're looking up by Id as we should, i don't think that should be an issue).

backend/routes/door.go
AdrianJendo: Again, i'm sending the entire object from the frontend so the payload should contain all data you need (including Id). I think put is simpler here even if we aren't updated every field. I'm not going to argue if you want to use PATCH but from what I've read PUT is a simpler option less prone to mistakes.

backend/controller/door.go
ABarroso647: I already looked into readonly fields, it doesn't work for id or CreatedAt as far as I can tell, because CreatedAt will not be updated on post if it is read only and making id readonly breaks.

Couple of other things, a lot of this is just like limitations with Golang itself, like in most languages you can support generic interfaces that all have a common subset and its just weird in go. 

Also even if I removed common entirely here,  the only thing that would change is you wouldn't have a pointer. Esssentially the pointer is saying whenever you update door (via copying the input values to the found door object), also update "updatedCommon" (see helper function) so that the static "initCommon" can be compared to see if there were any common values. The main issue is the interface we have "T" will not let us do T.Common, even though every object within T contains common, there's no real golang workaround for that. However I will say, this pointer method once u like take 2 seconds thinking abt pointers does make sense and I can reuse this to handle all future issues with Common comparisons as this has been my major issue how to generically pass a common function that is changing.

I haven't looked into embedded, didn't see anything written on it when I was looking through solutions to our problems so I also don't think it'll help because we are trying to solve a typing issue not a GORM issue, but I can see.

I could in theory bring back the "reqUser" and "reqDoors" I did see a json marshal, unmarshal method that may be promising to go from reqUser->User this way reqUser will only contain the editable fields, but it does add back in a decent amount of extra boilerplate hence why I didn't immediately do it when I had this working.

backend/models/user.go
ABarroso647: It would only change if its True, and you don't pass anything it would assume u want to make it false

ABarroso647: Also, I think we should put the default under gorm instead of JSON, this would keep readability and then deal with that issue aswell. 
image

backend/helpers/helpers.go
ABarroso647: one is a pointer to the original object, which was updated and one isn't a pointer

backend/helpers/helpers.go
ABarroso647: mb, left in from debugging

backend/models/user.go
AdrianJendo: that works

backend/controller/door.go
AdrianJendo: I find it hard to believe gorm doesn't support a better way to do this, especially because our relations are not particularly complicated.

At least we should abstract the common stuff away from whoever is using these functions as much as possible, I don't see why we can't implement this logic in the function body instead of ask for the user to pass in more parameters.

cameronkinsella: I agree that updating Common is making this a bit complicated when it was initially supposed to do the opposite.

Do we actually need `CreatedAt` or `UpdatedAt` anywhere? If there's no value in keeping them then I think we should get rid of Common to make all of this more readable.

backend/helpers/helpers.go
```diff
@@ -72,9 +73,27 @@ func GetFirstTable[T models.AllTables, P models.AllTables](w http.ResponseWriter
 
 }
 
+func UpdateSpecifiedParams[T models.AllTables](w http.ResponseWriter, r *http.Request, table *T, initCommon models.Common, updatedCommon *models.Common) (error, T) {
+	err := json.NewDecoder(r.Body).Decode(&table)
+	if err != nil {
+		http.Error(w, "400 malformed request", http.StatusBadRequest)
```
cameronkinsella: nit: I don't think the status code should be included in the error message.

backend/controller/door.go
ABarroso647: not sure about usage, I do know that comparing id is going to be the exact same as comparing common (generically) cuz essentially this whole comparing commons still non generically would boil down to 1 if statement. Unless you know of some way i don't to compare it

backend/helpers/helpers.go
ABarroso647: if @AdrianJendo cares I'll update them all

backend/routes/door.go
cameronkinsella: I don't have a strong opinion on this. Whatever is easiest to implement on the backend. I would assume that PUT would be easier to handle, but if it makes no difference then its fine.

backend/helpers/helpers.go
cameronkinsella: Every usage I see uses `foo` and `&foo`, the same variable before any modification. In what scenario could they be different?

backend/controller/door.go
cameronkinsella: Ah, you're right.
Maybe it's still worth doing if we're not using `CreatedAt` or `UpdatedAt`. Then we won't need to worry about if we should/shouldn't update them.
I guess adding a method to get `id` in the interface might be worthwhile if you're using it a lot.

ABarroso647: if that works lol, tried doing something similar with common

ABarroso647: could be mb tho ig

backend/helpers/helpers.go
ABarroso647: image
in this scenario, cuz we do a search for backend by id, that goes into "foo", and then we pass by value foo.common, but also pass by reference foo.common, and pass by reference foo. In this case when we use the line below:
err := json.NewDecoder(r.Body).Decode(&table)

we are updating the table from the backend adding on top any values inputted by the user in the request body, this will also update the pass by reference foo.common but crucially not the pass by value one which retains the original values returned from the db. Now in the case of the image above, we would have a mismatch where the user added an id in the body which doesn't match the id of the object we are trying to update. (you can imagine why this would cause issues, esp cuz gorm will automatically insert if the id doesn't exist iirc). As such we do this comparison bass by reference foo doesn't equal pass by value food and we know there's been an invalid change to uneditable data.

backend/data/lynx-web.sample.db
ABarroso647, line 1: here is your sample db, it has 103 users, the first 3 being the main ones, with  key/s (i think only the 1st user has multiple keys), role associations, with each of the role associations being associated with door/s

ABarroso647: image
users
image
roles
image
doors
image
sample result showing access for the role: janitor 
image
sample result showing roles for user "admin"
image
sample result for user with multiple roles, this is user clean cammy

backend/models/credential.go
ABarroso647, line 1: I genuinely didn't see purpose of leaving id here, everything seems to still work, so lmk

ABarroso647: it also wasn't acting as a pk for me, idk if it was when we had it as common so they were all 0

AdrianJendo: well WebauthnId has the primaryKey annotation so taht's why. I'm not sure if I added that by mistake. Doesn't really make a difference, but I still think an integer id is better if we're dealing with relations. Just makes it more intuitive.

AdrianJendo: If we want to leave out Id, i think we can change webauthnId to Id, would just require us to refactor is uses on the FE / BE. If that's what we want to do, then just open up a new PR off this branch and rename WebauthnId to Id

AdrianJendo (commented):
lgtm, just lmk what your thoughts are about the one comment.

backend/models/credential.go
ABarroso647: Im fine with renaming it, I thought there was a reason we had to keep it as webauthnid cuz of the serializer to json. But if we don't have to setting it as an id will be easier to refer to them

ABarroso647: oh i see u mean just rename it and keep everything else the same, yea i fw that. can do that

AdrianJendo: Ok let's do that in a future PR

AdrianJendo (approved):
Just note the comment about updating webauthnid in a future PR

cameronkinsella (approved):

PR #29, opened 2024-03-04
Frontend table updates

AdrianJendo:
https://github.com/lynx-locks/lynx-web/assets/55325093/963cb38d-e624-48bc-8bad-e8cd99e07e0c

There's going to be a lot to implement on the frontend for adding backend functionality, just going to be splitting it up as best I can :)


backend/controller/login.go
```diff
@@ -5,11 +5,10 @@ import (
 	"encoding/json"
 	"github.com/golang-jwt/jwt/v5"
 	"net/http"
-	"os"
 	"time"
 )
 
-var secretKey = []byte(os.Getenv("JWT_SECRET"))
+var secretKey = []byte("secret")
```
AdrianJendo, line 11: this wasn't supposed to be changed in the first places oopsies

frontend/src/components/adminTable/table.tsx
```diff
@@ -63,39 +143,62 @@ export default function AdminTable({ users }: { users: User[] }) {
...
+                  placeholder="Select Role..."
+                  subheader=""
+                  isMulti
                 />
               </div>
               <div className={styles.settingsButtonGroup}>
-                <button className={styles.settingsButton}>Update role</button>
                 <button className={styles.settingsButton}>
-                  Revoke & Regenerate credentials
+                  Revoke credentials
```
cameronkinsella: nit: `Credentials` should maybe be capitalized here for consistency. You seem to capitalize the first letter in each word for every other button.

frontend/src/components/dashboard/dashboard.tsx
```diff
@@ -3,59 +3,60 @@ import Search from "@/components/search/search";
...
+
   useEffect(() => {
-    const f = async () => {
-      const users = axios.get("/users");
+    const getUsers = async () => {
+      const users = await axios.get("/users");
+      setUsers(
+        users.data.map((user: User) => {
+          user.timeIn = new Date().toLocaleTimeString(); // TODO: parse these from response
+          user.lastDateIn = new Date().toLocaleDateString();
```
cameronkinsella: Docker containers use UTC by default, so this will not be real local time.
I think it's fine to leave for this PR, but can you please put a TODO here?
```diff
        users.data.map((user: User) => {
+          // TODO obtain timezone from TZ environment variable
          user.timeIn = new Date().toLocaleTimeString(); // TODO: parse these from response
          user.lastDateIn = new Date().toLocaleDateString();
```

AdrianJendo: Sure, ill just add it onto the TODO that's already there. Basically these values should be stored on the user object. I guess if we store it as a unix timestamp (which we should), we still need to do the timezone logic.

cameronkinsella (dismissed):

cameronkinsella (approved):

PR #35, opened 2024-03-09
Add UI / payload for registering yubikey

AdrianJendo:
Screenshot 2024-03-09 at 3 42 53 PM
Screenshot 2024-03-09 at 3 42 45 PM
Screenshot 2024-03-09 at 3 42 34 PM

### DIFF
--- a/frontend/src/app/register/page.tsx
+++ b/frontend/src/app/register/page.tsx
@@ -58,7 +58,7 @@ export default function RegisterUser() {
           <input
             className={styles.textInput}
             type="text"
-            placeholder="Yubikey Serial Number (leave blank if not familiar to you)"
+            placeholder="Yubikey Serial Number (leave blank if unfamiliar to you)"
             value={yubikeySerial}
             onChange={(e) => setYubikeySerial(e.target.value)}
           />
--- a/frontend/src/app/register/page.tsx
+++ b/frontend/src/app/register/page.tsx
@@ -10,6 +10,32 @@ import LoadingStatus from "@/types/loadingStatus";
 import styles from "./register.module.css";
 import Loader from "@/components/loader/loader";
 
+const modhex = {
+  c: "0",
+  b: "1",
+  d: "2",
+  e: "3",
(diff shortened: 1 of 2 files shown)


ABarroso647:
LGTM: @cameronkinsella  give final say on this

frontend/src/app/register/page.tsx
```diff
@@ -45,10 +48,21 @@ export default function RegisterUser() {
...
+          <div className={styles.registerButton}>
+            <SubmitButton
+              onClick={() => registerPasskey()}
+              text="Create a New Passkey"
+            />
+          </div>
+          <input
+            className={styles.textInput}
+            type="text"
+            placeholder="Yubikey Serial Number (leave blank if unfamiliar to you)"
```
cameronkinsella: This will be the YubiKey OTP, not the serial number/yubikey id itself. We are extracting the Yubikey ID from the OTP.
YubiKey users are generally familiar with how to use their OTP, but often do not know their YubiKey ID.

```diff
-            placeholder="Yubikey Serial Number (leave blank if unfamiliar to you)"
+            placeholder="Paste YubiKey OTP (leave blank for passkeys)"

cameronkinsella (changes requested):
In addition to my comment, I think it would be more secure to parse the YubiKey ID from the OTP on the frontend instead of the backend. It's just the first 12 characters of the OTP that the user enters.

You can also do some error checking on the frontend to alert the user before POSTing, since we know a valid OTP is always 44 characters total.

[More info on YubiKey OTP](https://developers.yubico.com/OTP/OTPs_Explained.html)

frontend/src/app/register/page.tsx
cameronkinsella: Also don't forget that "YubiKey" has a capital "K".

cameronkinsella (changes requested):
Just add the conversion to decimal here and it should be good to merge.

cameronkinsella (approved):

PR #38, opened 2024-03-11
Update backend workflow such that frontend can run in full and door unlock testing can be done

ABarroso647:
Changes include:
-Updating association to be user->roles
-Updating user post request to allow the addition of roles on creation example request shown below:

-Updating the user put request to allow the updating of both roles and user fields at the same time with one reuqest, example request shown below:

-Adding getAllkeys function to get all keys for a user

-Adding delete propogation such that, when deleting associations are automatically deleted and in the case of many to one, like our users, all associated keys are deleted upon user delete

-Updated role post similarily to users to allow door associations to be added on create.

-Adding in full workflow to check if user has access to the door upon login, and then to unlock door (mobile flow),
-Bugfix on login logic

### DIFF
--- a/backend/config/db.go
+++ b/backend/config/db.go
@@ -16,7 +16,12 @@ func Connect() {
 	if err != nil {
 		panic("failed to create data directory")
 	}
-	db, err := gorm.Open(sqlite.Open("data/lynx-web.db"), &gorm.Config{TranslateError: true, Logger: logger.Default.LogMode(logger.Info)})
+	db := &gorm.DB{}
+	if value, ok := os.LookupEnv("NODE_ENV"); ok && value == "production" {
+		db, err = gorm.Open(sqlite.Open("data/lynx-web.db"), &gorm.Config{TranslateError: true})
+	} else {
+		db, err = gorm.Open(sqlite.Open("data/lynx-web.db"), &gorm.Config{TranslateError: true, Logger: logger.Default.LogMode(logger.Info)})
+	}
 	if err != nil {
 		panic("failed to connect database")
 	}
--- a/backend/config/db.go
+++ b/backend/config/db.go
@@ -5,6 +5,7 @@ import (
 	"api/models"
 	"gorm.io/driver/sqlite"
 	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
 	"os"
 	"path/filepath"
 )
@@ -15,7 +16,7 @@ func Connect() {
 	if err != nil {
 		panic("failed to create data directory")
 	}
-	db, err := gorm.Open(sqlite.Open("data/lynx-web.db"), &gorm.Config{TranslateError: true})
+	db, err := gorm.Open(sqlite.Open("data/lynx-web.db"), &gorm.Config{TranslateError: true, Logger: logger.Default.LogMode(logger.Info)})
 	if err != nil {
 		panic("failed to connect database")
 	}
--- a/backend/controller/role.go
+++ b/backend/controller/role.go
@@ -41,13 +41,19 @@ func CreateRole(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		http.Error(w, "Malformed request", http.StatusBadRequest)
 	}
-	doors := role.Doors
-	for i, _ := range doors {
-		res := db.DB.First(&doors[i])
+	ids := helpers.GetAllIdsFromList(role.Doors)
+	if len(ids) != 0 {
+		doors := []models.Door{}
+		res := db.DB.Find(&doors, ids)
 		if res.Error != nil {
-			http.Error(w, "One or more invalid doors entered", http.StatusBadRequest)
+			helpers.DBErrorHandling(res.Error, w)
 			return
 		}
+		if len(doors) != len(ids) {
+			http.Error(w, "One or more invalid roles", http.StatusBadRequest)
+			return
+		}
+		role.Doors = doors
 	}
 	err, role = helpers.CreateNewRecord(w, role)
 	if err != nil {
--- a/backend/controller/role.go
+++ b/backend/controller/role.go
@@ -41,6 +41,14 @@ func CreateRole(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		http.Error(w, "Malformed request", http.StatusBadRequest)
 	}
+	doors := role.Doors
+	for i, _ := range doors {
+		res := db.DB.First(&doors[i])
+		if res.Error != nil {
+			http.Error(w, "One or more invalid doors entered", http.StatusBadRequest)
+			return
+		}
+	}
 	err, role = helpers.CreateNewRecord(w, role)
 	if err != nil {
 		return
@@ -54,13 +62,8 @@ func DeleteRole(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		return
 	}
-	err = helpers.DeleteByPk(w, models.Role{}, rId)
+	err = helpers.DeleteObjandAssociationsByPk(w, models.Role{Id: rId})
 	if err != nil {
-		return
-	}
-	err = db.DB.Unscoped().Model(&models.Role{Id: rId}).Association("Doors").Unscoped().Clear()
-	if err != nil {
-		http.Error(w, "500 unable to delete association/s", http.StatusInternalServerError)
 		return
 	}
 	w.WriteHeader(200)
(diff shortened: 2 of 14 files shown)


backend/config/db.go
```diff
@@ -15,7 +16,7 @@ func Connect() {
 	if err != nil {
 		panic("failed to create data directory")
 	}
-	db, err := gorm.Open(sqlite.Open("data/lynx-web.db"), &gorm.Config{TranslateError: true})
+	db, err := gorm.Open(sqlite.Open("data/lynx-web.db"), &gorm.Config{TranslateError: true, Logger: logger.Default.LogMode(logger.Info)})
```
AdrianJendo: what for?

backend/controller/role.go
```diff
@@ -41,6 +41,14 @@ func CreateRole(w http.ResponseWriter, r *http.Request) {
...
 		http.Error(w, "Malformed request", http.StatusBadRequest)
 	}
+	doors := role.Doors
+	for i, _ := range doors {
+		res := db.DB.First(&doors[i])
+		if res.Error != nil {
+			http.Error(w, "One or more invalid doors entered", http.StatusBadRequest)
+			return
+		}
+	}
```
AdrianJendo: This seems wildly inefficient, can we do something like this and compare the lengths:

Screenshot 2024-03-11 at 3 38 30 PM
Screenshot 2024-03-11 at 3 39 10 PM
Screenshot 2024-03-11 at 3 39 01 PM
Screenshot 2024-03-11 at 3 38 50 PM

Reference: https://gorm.io/docs/query.html#Retrieving-all-objects

backend/controller/user.go
```diff
@@ -34,15 +34,27 @@ func UpdateUser(w http.ResponseWriter, r *http.Request) {
...
 	}
+	roles := user.Roles
+	for i, _ := range roles {
+		res := db.DB.First(&roles[i])
+		if res.Error != nil {
+			http.Error(w, "One or more invalid roles entered", http.StatusBadRequest)
+			return
+		}
+	}
 	err, user = helpers.UpdateObject(w, user)
```
AdrianJendo: Same here, let's not do n separate SELECT statements when we can just do 1.

backend/controller/user.go
```diff
@@ -34,15 +34,27 @@ func UpdateUser(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
 		return
 	}
+	err = db.DB.Debug().Model(&user).Association("Roles").Replace(&roles)
```
AdrianJendo: .Debug

backend/controller/user.go
```diff
@@ -67,9 +79,18 @@ func GetUserByEmail(w http.ResponseWriter, r *http.Request) {
...
+
 	err := json.NewDecoder(r.Body).Decode(&user)
+	roles := user.Roles
+	for i, _ := range roles {
+		res := db.DB.First(&roles[i])
+		if res.Error != nil {
+			http.Error(w, "One or more invalid roles entered", http.StatusBadRequest)
+			return
+		}
+	}
```
AdrianJendo: same here

backend/controller/webAuthn.go
```diff
@@ -155,8 +156,7 @@ func RegisterResponse(w http.ResponseWriter, r *http.Request) {
 	// Add public key to DB
 	dbCredential := models.Credential{
 		UserId:          user.Id,
-		Roles:           nil, // TODO: use activeTokens table to find roles
-		WebauthnId:      credential.ID,
+		Id:              credential.ID,
```
AdrianJendo, line 159: Let's move this to the first row in the struct.

backend/controller/webAuthn.go
```diff
@@ -266,15 +268,44 @@ func AuthorizeResponse(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
-
-	err = SaveDBCredential(w, credential)
-	if err != nil {
+	//err = SaveDBCredential(w, credential)
+	//if err != nil {
+	//	http.Error(w, "Could not finish log in", http.StatusInternalServerError)
+	//	return
+	//}
```
AdrianJendo, line 280: Let's add a comment explaining what this is and why it was commented out.

backend/controller/webAuthn.go
```diff
@@ -266,15 +268,44 @@ func AuthorizeResponse(w http.ResponseWriter, r *http.Request) {
...
+
+	user := models.User{}
+	creds := models.Credential{Id: credential.ID}
+	res := db.DB.First(&creds)
+	if res.Error != nil {
+		http.Error(w, "Could retrieve credentials ", http.StatusInternalServerError)
```
AdrianJendo: "Could not retrieve credential"

backend/controller/webAuthn.go
```diff
@@ -266,15 +268,44 @@ func AuthorizeResponse(w http.ResponseWriter, r *http.Request) {
...
+	db.DB.Model(&creds).Association("User").Find(&user)
+	roles := []models.Role{}
+	db.DB.Model(&user).Association("Roles").Find(&roles)
+	doors := []models.Door{}
+	db.DB.Distinct("id").Model(&roles).Association("Doors").Find(&doors)
+	hasAccess := false
+	for _, curDoor := range doors {
+		if curDoor.Id == dId {
+			hasAccess = true
+		}
```
AdrianJendo: Let's break if we find the id

backend/controller/webAuthn.go
```diff
@@ -266,15 +268,44 @@ func AuthorizeResponse(w http.ResponseWriter, r *http.Request) {
...
+	}
+	if hasAccess == false {
```
AdrianJendo: `!hasAccess`

backend/controller/webAuthn.go
AdrianJendo: Actually, can we just use a built-in search function: https://stackoverflow.com/questions/38654383/how-to-search-for-an-element-in-a-golang-slice
https://pkg.go.dev/slices#IndexFunc

backend/controller/webAuthn.go
```diff
@@ -266,15 +268,44 @@ func AuthorizeResponse(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, "User does not have access to door", http.StatusBadRequest)
+		return
+	}
+	errJson := json.NewEncoder(w).Encode("Login Success, opening door")
 	if errJson != nil {
 		http.Error(w, "Could not return results", http.StatusInternalServerError)
 		return
 	}
+	go unlockDoor(dId)
```
AdrianJendo: Why do you write `go unlockDoor()` instead of `unlockDoor()`

AdrianJendo:
```
x := slices.IndexFunc(doors, func(d modes.Door) bool { return d.Id == dId }
if x == -1 {
  // error logic
  return
}

// success logic

```

backend/routes/door.go
```diff
@@ -11,5 +11,6 @@ func DoorsRoute() chi.Router {
 	r.Post("/", controller.CreateDoor)
 	r.Put("/", controller.UpdateDoor)
 	r.Delete("/{doorId}", controller.DeleteDoor)
+	r.Get("/unlocked/{doorId}", controller.OpenDoor)
```
AdrianJendo: `unlocked` or `unlock`?

backend/routes/user.go
```diff
@@ -13,7 +13,7 @@ func UsersRoute() chi.Router {
 	r.Put("/", controller.UpdateUser)
 	r.Post("/", controller.CreateUser)
 	r.Delete("/{userId}", controller.DeleteUser)
-	r.Post("/{userId}/roles", controller.ReplaceRoleAssociations)
 	r.Get("/{userId}/roles", controller.GetRoleAssociations)
+	r.Get("/{userId}/creds", controller.GetCredAssociations)
```
AdrianJendo: Can we call these like `GetUserRoles` and `GetUserCreds` or something

AdrianJendo (commented):
I don't think this should all be in 1 PR for the future

backend/controller/webAuthn.go
ABarroso647: what does this refer to?

backend/routes/door.go
ABarroso647: well a door is "unlocked" not unlock

AdrianJendo: What does this endpoint do?

ABarroso647: this is the endpoint cameron is pinging, so it tells him when a door has been "unlocked" by a user so he can run the logic to open it

AdrianJendo: Can you change the controller name this to `IsDoorOpen` or something

AdrianJendo (approved):

PR #39, opened 2024-03-12
Add in registration flow with email

ABarroso647:
https://github.com/lynx-locks/lynx-web/assets/57304476/bf8f03ce-4ba5-475c-b939-abd25f64e40c

Also adding in an env variable you will need to add to this for this feature


.sample.env
```diff
@@ -0,0 +1,3 @@
+CLIENT_DOMAIN="your production website here"
+HOST_EMAIL_ADDRESS="your email here"
+HOST_EMAIL_PASSWORD="your email password here"
```
AdrianJendo: can u call it .template.env

AdrianJendo (approved):

PR #40, opened 2024-03-13
Update frontend to support auth

AdrianJendo:
- Frontend logic changes to adapt to camerons auth updates in #44


backend/controller/user.go
```diff
@@ -62,6 +67,12 @@ func UpdateUser(w http.ResponseWriter, r *http.Request) {
 		}
 		user.Roles = roles
 	}
+	if reqUser.Email != "" {
```
ABarroso647, line 70: this is just not a put request anymore?

backend/controller/user.go
```diff
@@ -197,6 +208,7 @@ func SendRegistrationEmail(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 	}
+	w.WriteHeader(200)
```
ABarroso647, line 211: it already sends 200, but im fine with keeping this since we do it elsewhere just maybe we don't need to be doing it elsewhere

backend/controller/user.go
AdrianJendo: we're not storing data such as the webauthnid on the frontend so i can't send all the user fields easily

ABarroso647: also this is fine for now but if we genuinely want a patch req here, I want to make a note that I'm gonna rewrite this logic

ABarroso647: cuz this is ugly asf

cameronkinsella (dismissed):

cameronkinsella (approved):

PR #43, opened 2024-03-13
Add tracking for last time a user opened a door

ABarroso647:


backend/controller/door.go
```diff
@@ -67,11 +67,17 @@ func IsDoorOpen(w http.ResponseWriter, r *http.Request) {
...
+	uId, hasAccess := models.DoorUnlocked[dId]
+	if !hasAccess {
 		http.Error(w, "Door not unlocked", http.StatusUnauthorized)
+		return
+	}
+
+	err = UpdateLastTimeIn(w, uId)
+	if err != nil {
+		http.Error(w, "Unable to update last time in", http.StatusInternalServerError)
+		return
```
AdrianJendo, line 79: I'm confused I thought this was the endpoint that was constantly pinged by the door? Does this logic belong here?

backend/controller/webAuthn.go
```diff
@@ -345,14 +346,22 @@ func GetYubiKeyDoorAccess(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
-	if !credHasDoorAccess(cred, dId) {
+	hasAccess, user := credHasDoorAccess(cred, dId)
+	if !(hasAccess) {
 		http.Error(w, "User does not have access to door", http.StatusBadRequest)
 		return
 	}
+
+	err = UpdateLastTimeIn(w, user.Id)
```
AdrianJendo, line 355: Isn't this the only place we need this logic?

backend/controller/door.go
ABarroso647: it won't ever reach that step, this is when the user unlocks the door so yes why would this not be where the logic belongs? they unlock door on mobile we update their time in, otherwise we don't update it

backend/controller/webAuthn.go
ABarroso647: no this is when they open the door with a yubikey, since there are two flows we have two spots

AdrianJendo (approved):

PR #44, opened 2024-03-14
Route JWT Auth Protection

cameronkinsella:
Implemented authentication middleware that can be added to routes. The current groups are:
- Admin only
- User only
- Public

Only the `/admin` and `/` routes are currently protected. We should later make most of our API routes protected as well, but that isn't necessary for the demo.

https://github.com/lynx-locks/lynx-web/assets/53635597/134efdeb-f107-425f-ba0a-5bcc0835c224


backend/controller/login.go
```diff
@@ -78,59 +59,80 @@ func LoginResponse(w http.ResponseWriter, r *http.Request) {
...
-			http.Error(w, "Could not return results", http.StatusInternalServerError)
-			return
-		}
-	} else {
-		http.Error(w, "User not admin", http.StatusUnauthorized)
+	// Get the session data stored from the function above
+	sessionData := models.SessionData{
+		UserId: user.WebAuthnID(),
+	}
+	result := db.DB.First(&sessionData)
```
ABarroso647, line 66: did you verify sql for this line, if you see other places we are querying by webauthnID we use take example in RegisterResponse(): 	result := db.DB.Where(models.SessionData{UserId: user.WebAuthnID()}).Take(&sessionData)

AdrianJendo: This is getting refactored in my PR @cameronkinsella you can just copy / paste these lines:
https://github.com/lynx-locks/lynx-web/blob/7392f230d61794884b003bc0e707c03e9ad4ab23/backend/controller/login.go#L53-L89

backend/server.go
```diff
@@ -76,7 +87,30 @@ func main() {
...
 	config.Connect()
-	FileServer(r, "/", http.Dir("./static"))
+
+	// Admin only
+	r.Group(func(r chi.Router) {
+		// Seek, verify and validate JWT tokens
+		r.Use(jwtauth.Verifier(auth.TokenAuth))
+		r.Use(auth.VerifyAdmin)
+
+		ServeFrontendRoute(r, "/admin/")
```
AdrianJendo, line 97: is the extra slash important?

frontend/src/app/admin/page.tsx
```diff
@@ -0,0 +1,21 @@
...
+    setName(storedName);
+  }, []);
+
+  return (
+    <div>
+      <Navbar name={name} />
+      <Dashboard />
+    </div>
+  );
+}
```
AdrianJendo: nice, i'd prefer to store the whole user instead of just the name using `JSON.stringify` and `JSON.parse` in the local storage, but I suppose I can change that after

backend/controller/login.go
cameronkinsella: This was not modified. This diff is a bit misleading because this section is no longer under the `if user.IsAdmin` branch (since this login works for both users and admin).

frontend/src/app/login/page.tsx
```diff
@@ -47,21 +24,23 @@ const Login = () => {
...
         ...credential,
         challenge: options.challenge,
       });
 
-      const { token } = verifyResp.data;
-      if (verifyResp.status === 200 && token) {
-        localStorage.setItem("token", token);
+      if (verifyResp.status === 200) {
+        // We can set any non-private user info in local storage to avoid addition requests
+        localStorage.setItem("name", user.name);
```
AdrianJendo, line 39:
```suggestion
        localStorage.setItem("user", JSON.stringify(user);
```

backend/controller/login.go
ABarroso647: i see, that is probably still a bug tho considering it was a bug on register

frontend/src/app/admin/page.tsx
```diff
@@ -0,0 +1,21 @@
...
+import Navbar from "@/components/navbar/navbar";
+import { useEffect, useState } from "react";
+
+export default function Home() {
+  const [name, setName] = useState<string>("");
+  useEffect(() => {
+    const storedName = localStorage.getItem("name")
+      ? String(localStorage.getItem("name"))
+      : "";
+    setName(storedName);
```
AdrianJendo:
```suggestion
    if (localStorage.getItem("user") {
        setName(JSON.parse(localStorage.getItem("user")).name)
    }
```

This might not work off the bat but something like this

frontend/src/app/page.tsx
```diff
@@ -1,51 +1,20 @@
...
-        console.error("Token error: ", error);
-        localStorage.removeItem("token");
-      }
-    }
-    // Redirect to login
-    redirect("/login/?referrer=/");
+    const storedName = localStorage.getItem("name")
+      ? String(localStorage.getItem("name"))
+      : "";
+    setName(storedName);
```
AdrianJendo: user

frontend/src/app/admin/page.tsx
cameronkinsella: local storage is insecure and `user` contains the user's email which can be seen as private information, so I avoided storing it.
Up to you if you want to update it.

AdrianJendo: I felt like wanting to know the user's email or id might be useful but i guess it's not a big deal

backend/controller/login.go
AdrianJendo: It is modified @cameronkinsella. Specifically the part adam commented on

AdrianJendo: I can add it if that's easier

frontend/src/app/admin/page.tsx
cameronkinsella: I think `id` is fine to store if we need it, but I didn't see it used anywhere so I did not store it in this PR.
We can always get the user's email through `GET /api/user/{id}`, so I think we should avoid putting it in insecure storage just to avoid the extra request.

AdrianJendo: ok

backend/server.go
cameronkinsella: No

AdrianJendo: Why do we have it then, can this be changed to `/admin`?

cameronkinsella: oops, misread what this was commenting on.
Yes, this is important. the frontend routing needs to end with a `/` for the route.

AdrianJendo: Why, what happens if we omit both the FE and BE trailing slash?

ABarroso647 (approved):

backend/server.go
cameronkinsella: If you omit the training slash, the routing still works but the auth middleware does not catch it. So it goes through unauthenticated.
This behavior is only for the FE I believe. It's the result of file server routing in `chi` being not as flexible as normal endpoints.

AdrianJendo (approved):

PR #45, opened 2024-03-14
Add ability to delete all of a user's creds

ABarroso647:


backend/controller/user.go
```diff
@@ -165,6 +166,24 @@ func GetUserCreds(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func DeleteUserCreds(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	err, uId := helpers.ParseInt(w, r, "userId")
+	if err != nil {
+		return
+	}
+	creds := []models.Credential{}
+	res := db.DB.Unscoped().Select(clause.Associations).Where(models.Credential{UserId: uId}).Delete(&creds)
```
AdrianJendo:
```suggestion
	res := db.DB.Unscoped().Select(clause.Associations).Where(models.Credential{UserId: uId}).Delete(&[]models.Credential{})
```

AdrianJendo (dismissed):
Otherwise lgtm

backend/controller/user.go
```diff
@@ -165,6 +166,24 @@ func GetUserCreds(w http.ResponseWriter, r *http.Request) {
...
+	if res.Error != nil {
+		helpers.DBErrorHandling(res.Error, w)
+		return
+	}
+	if res.RowsAffected < 1 {
+		http.Error(w, "No credentials found, either user has none or user does not exist", http.StatusNotFound)
+		return
+	}
+}
+
```
AdrianJendo: Also as a nit, I'm not a huge fan of returning nothing, it makes working with the API a bit more annoying. I think adding the `WriteHeader(200)` logic was good as a placeholder. There really isn't a need to add it at this point but i think its a good practice.

AdrianJendo (approved):

PR #48, opened 2024-03-14
Implement a hopefully bug fix

AdrianJendo:
This might not fix the bug, but I think our system is at least more robust this way. There are legitimate cases where a user might start the login process, but if they begin the login and don't finish (ie: quit before hitting the login/auth/register response endpoint), then their session should be persisted in the backend (and that is expected). Thus, we need a way to clear out stale sessions. Having a refresh button that hits an endpoint @ABarroso647 will create to delete the entire session table is ok (we can say it's simulating a cron job). But changing the SQL query as follows is also a good step. This prevents one user from having more than one active session (which is fine), and allows us to resync their session data upon desynchronization (ie: failing to hit the `/response` endpoint):

These screenshots are from registering a new key:
Before:
Screenshot 2024-03-14 at 5 28 28 PM

After:
Screenshot 2024-03-14 at 5 33 27 PM

As can be seen, the session is deleted by user id, meaning that if a session gets de-synced and not deleted, the user's next registration attempt might fail (by picking the wrong session), but at least both of those users sessions will get deleted (allowing their next re-attempt to succeed).

This mainly affects flows where the user is known (login & register). The auth flow should be unaffected because it indexes by challenge (and Jonah's problems were with the registration, so I think we can probably assume authorization is fine for now).

PS: I think the `ActiveTokens` table works fine since it indexes by the token id in the email, so there would only ever be 1 entry created in the `SendEmail` endpoint, and this token would only be deleted when the user successfully registers.

### DIFF
--- a/frontend/src/components/buttonsRow/buttonsRow.tsx
+++ b/frontend/src/components/buttonsRow/buttonsRow.tsx
@@ -50,6 +50,7 @@ export default function ButtonRow({
       id: 1,
       name: "New User",
       onClick: async () => {
+        setDisabled(false);
         setRoles(await getRoleOptions());
         setNewUserModal(true);
       },
@@ -57,12 +58,16 @@ export default function ButtonRow({
     {
       id: 2,
       name: "New Key",
-      onClick: () => setNewKeyModal(true),
+      onClick: () => {
+        setDisabled(false);
+        setNewKeyModal(true);
+      },
     },
     {
       id: 3,
       name: "New Role",
       onClick: async () => {
+        setDisabled(false);
         setDoors(await getDoorOptions());
         setNewRoleModal(true);
       },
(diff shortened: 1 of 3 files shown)


frontend/src/components/buttonsRow/buttonsRow.tsx
```diff
@@ -50,19 +50,24 @@ export default function ButtonRow({
...
+      onClick: () => {
+        setDisabled(false);
+        setNewKeyModal(true);
+      },
     },
     {
       id: 3,
       name: "New Role",
       onClick: async () => {
+        setDisabled(false);
```
AdrianJendo, line 70: This is unrelated but i don't want to open another PR.

cameronkinsella (approved):

PR #50, opened 2024-03-15
Implement 401 denied page

AdrianJendo:
- @cameronkinsella can you hook this up to the backend so that instead of the "access denied - user not admin" page, they are redirected here?

Screenshot 2024-03-15 at 7 00 57 PM


cameronkinsella:
Added the redirect and tested.

AdrianJendo:
LGTM

ABarroso647 (approved):

PR #52, opened 2024-03-16
Update Python Scripts

AdrianJendo:
- Update python scripts to conform to API endpoints and to provide easy setup for testing


backend/models/door.go
```diff
@@ -2,7 +2,7 @@ package models
 
 type Door struct {
 	Id          uint   `json:"id"`
-	Name        string `json:"name"`
+	Name        string `json:"name" gorm:"unique"`
```
ABarroso647, line 5: ignore, considered it some more and im fine with this

ABarroso647 (approved):

PR #53, opened 2024-03-16
Update API responses

AdrianJendo:
- Update API responses to be consistent


backend/helpers/helpers.go
```diff
@@ -53,6 +53,18 @@ func JsonWriter(w http.ResponseWriter, table interface{}) {
 	}
 }
 
+func JsonWriterString(w http.ResponseWriter, message string) {
+	messageStruct := struct {
```
ABarroso647: may be a way to make this one function just by type checking if it's a string, potentially "reflect.Typeof()"

AdrianJendo: I just thought that would be weird since the parameter of the other one is called table

ABarroso647: i mean we can just rename it, I'm not particularly a fan of "JsonWriterString" being a separate function who's purpose is to remove quotes anyways

AdrianJendo: Well it does a little more than that but maybe you can refactor it to your liking and cam can approve?

ABarroso647: that sounds good

backend/helpers/helpers.go
```diff
@@ -46,20 +47,16 @@ func DBErrorHandling(error error, w http.ResponseWriter) {
...
+			Message string `json:"message"`
+		}{
+			Message: response.(string),
+		}
 	}
-	errJson := json.NewEncoder(w).Encode(&messageStruct)
+	errJson := json.NewEncoder(w).Encode(&response)
 	if errJson != nil {
 		http.Error(w, "Unable to encode response", http.StatusInternalServerError)
 	}
```
cameronkinsella: Reflection is very slow and creates the potential for runtime errors (the type is evaluated dynamically).
Reflection in Go is generally discouraged unless it's necessary, so I don't think it's worth introducing that behavior to save a few lines.

ABarroso647: there is no potential for runtime errors here

cameronkinsella: I know, not in this case. I was speaking about reflection in general.
I just prefer to discourage unnecessary and inefficient dynamic behavior. This type of pattern is not very idiomatic for Go in my opinion.
I'll leave the final call to @AdrianJendo.

cameronkinsella (approved):

backend/helpers/helpers.go
```diff
@@ -46,8 +46,16 @@ func DBErrorHandling(error error, w http.ResponseWriter) {
...
+func JsonWriter(w http.ResponseWriter, response interface{}) {
+	switch response.(type) {
+	case string:
+		response = struct {
+			Message string `json:"message"`
+		}{
+			Message: response.(string),
+		}
+	}
+	errJson := json.NewEncoder(w).Encode(&response)
```
AdrianJendo, line 58: @ABarroso647 can you indent the case here?

So this only executes the response = struct { ... } logic when response is a string right? Otherwise, the function is unchanged.

ABarroso647: image
copied formatting directly from golang website: https://go.dev/tour/flowcontrol/9
but yes, the struct logic does not happen when the response is not a string

PR #58, opened 2024-03-16
Update email logic

AdrianJendo:
- Wanted to add some better email handling
- Removing capitalization from email

### DIFF
--- a/backend/helpers/helpers.go
+++ b/backend/helpers/helpers.go
@@ -59,5 +59,6 @@ func ParseInt(w http.ResponseWriter, r *http.Request, key string) (error, uint) 
 }
 
 func FormatEmail(email string) string {
-	return strings.Replace(strings.ToLower(email), ".", "", -1)
+	x := strings.Split(email, "@")
+	return strings.ToLower(strings.Replace(x[0], ".", "", -1) + "@" + x[1])
 }
--- a/backend/helpers/helpers.go
+++ b/backend/helpers/helpers.go
@@ -1,16 +1,13 @@
 package helpers
 
 import (
-	"api/db"
-	"api/models"
 	"encoding/json"
 	"errors"
 	"github.com/go-chi/chi/v5"
-	"gorm.io/gorm"
-	"gorm.io/gorm/clause"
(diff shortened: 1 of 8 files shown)


backend/helpers/helpers.go
```diff
@@ -140,10 +50,7 @@ func ParseInt(w http.ResponseWriter, r *http.Request, key string) (error, uint)
...
 
-func GetAllIdsFromList[T models.AllTables](structList []T) []uint {
-	var list []uint
-	for _, element := range structList {
-		list = append(list, element.GetId())
-	}
-	return list
+func FormatEmail(email string) string {
+	x := strings.Split(email, "@")
+	return strings.ToLower(strings.Replace(x[0], ".", "", -1) + "@" + x[1])
```
ABarroso647: im so confused so what is this going to do with my email a.barroso2001@gmail.com?

AdrianJendo: https://support.google.com/mail/answer/7436150?hl=en#:~:text=Adding%20dots%20doesn't%20change,own%20johnsmith%40gmail.com.

AdrianJendo: From my understand, dots aren't a real thing in email addresses but rather a convenience. Same with caps

ABarroso647: I just don't see why we should change what the user entered as their email, like caps is nice sure, but otherwise what benefit does removing the dot give us?

cameronkinsella: > https://support.google.com/mail/answer/7436150?hl=en#:~:text=Adding%20dots%20doesn't%20change,own%20johnsmith%40gmail.com.

The dot rule only applies to gmail and a few other providers. This is not a universal rule, so we should not be applying it here. We should use whatever the user inputted to not accidentally break something that uses custom routing rules.

Case-insensitivity is universal to all email servers though, so I think it's fine to keep that.

AdrianJendo: ok, ill fix that when I have time

cameronkinsella (approved):

PR #61, opened 2024-03-19
Create default admin if no users exist

cameronkinsella:
This functionality will be required to avoid lockout on initial startup when the api routes are protected by token auth.

If no users exist on startup, a new admin user is created and a fixed UUID registration token with no expiration is created for them.
While this is active, you will be automatically redirected to the default admin registration page when you go to the login page (no email is sent).

The initial admin's name and email must be set through environment variables. Those environment variables are only required on first launch, they can be removed afterward for security.

For security, this flow can **only** occur when the `users` table is empty to ensure that any users registering through this insecure flow will not gain access to registered user information. This does not protect against lockout if an admin deletes all other admins and then themselves while other non-admin users are registered.

image

https://github.com/lynx-locks/lynx-web/assets/53635597/92e82d31-88dd-4dd6-b3bd-030b4b11d8f8


ABarroso647:
LGTM @AdrianJendo  I'll give u final approval on this flow since you are more separated from it

backend/auth/tokenAuth.go
```diff
@@ -88,6 +92,23 @@ func VerifyNotLoggedIn(next http.Handler) http.Handler {
 	})
 }
 
+func InitialAdminCheck(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var count int64
+		db.DB.Model(&models.ActiveTokens{Id: uuid.Max}).Count(&count)
```
AdrianJendo, line 98: Is using uuid.Max a good approach? Why can't we just read in the environment variables like we did in the setup?

backend/server.go
```diff
@@ -229,3 +244,33 @@ func ServeFrontendRoute(r chi.Router, route string) {
...
+	} else {
+		panic("DEFAULT_ADMIN_EMAIL environment variable not set")
+	}
+
+	result := db.DB.Create(&user)
+	if result.Error != nil {
+		panic("Error creating initial admin user: " + result.Error.Error())
+	}
+
+	activeToken := models.ActiveTokens{Id: uuid.Max, UserId: user.Id, ExpiryDate: math.MaxInt64}
```
AdrianJendo, line 271: and use a random uuid here

backend/auth/tokenAuth.go
cameronkinsella: We could generate the initial admin's token based on a UUID passed in an environment variable, but I don't think there is any benefit. It would be more inconvenient for a user to set up and it would not improve security, since you get redirected to the correct token registration page anyway and this flow only happens when there is no user data in the system.

AdrianJendo: Im just saying why cant we use a random UUID?

ABarroso647: why do we care what the uuid is, there are no users, that being said I feel like we could totally just make it random no?

ABarroso647: actually on second thought, then u need to actually grab the uuid if its random so its probably more hassle for zero benefit

AdrianJendo: he selects by uuid later so he just needs to know if i guess

ABarroso647: yea but why tho, feels like extra work we don't need to do

ABarroso647: cuz if he sets it to a random token, he'd need to store what that random token is, ig in memory or smthn, in order for the select to check if that specific token is made.

cameronkinsella: > actually on second thought, then u need to actually grab the uuid if its random so its probably more hassle for zero benefit

Yeah, that's the idea. The benefit of a random UUID is security through its randomness, since a bad actor would need to guess a correct UUID to steal a user's registration. In the initial admin flow however, we're redirecting any user directly to the registration page so the value of the UUID does not matter. For this flow, the randomness has no benefit.

By using a fixed UUID we gain two benefits:
1. We make it more explicit that this "special-case" registration token has no expiration.
2. We can more easily redirect the user to the initial admin registration page without needing to find the correct UUID.

AdrianJendo (approved):
Fine, ill allow it

PR #62, opened 2024-03-19
Protect API endpoints with token auth middleware

cameronkinsella:
Protects most endpoints with token auth. The only endpoints not protected are the ones required for login, where the user does not yet have an auth token.

Since endpoints now require authentication, you will need to add a Bearer Token type Authorization header if you are testing endpoints in something like Postman. To obtain a token, you can:
1. Click the `Copy Token` button while logged in as an admin (dev or prod).
image
2. Copy the token that is created and logged when the backend server starts (**dev only**).
image
3. Copy the value of your `jwt` cookie while logged in as an admin (dev or prod)
image

The token auth verifier will check for both cookie or bearer token, so if you are already logged in as an admin then you can access GET endpoints just by visiting their URL.

The following frontend functionality has been tested and is not broken:
- Create user
- Delete user
- Add new key
- Revoke keys
- User registration
- User login/logout
- Create/Edit role
- Update user roles
- Door auth with passkey


cameronkinsella:
Added a settings modal. It can be opened by clicking on the cog icon on the top right. It currently only contains the logout option and (if the user is an admin) the option to copy their access token. Feel free to adjust the look and feel of these components.

I also added a disclaimer to educate the user about misuse of the access token. They must acknowledge this disclaimer before they are able to copy the token. The disclaimer is a bit long, so it becomes scrollable if the max height for it is reached. 

https://github.com/lynx-locks/lynx-web/assets/53635597/d4c54083-1455-419f-b80f-eb5684d3c1bb

AdrianJendo:
> Added a settings modal. It can be opened by clicking on the cog icon on the top right. It currently only contains the logout option and (if the user is an admin) the option to copy their access token. Feel free to adjust the look and feel of these components.
> 
> I also added a disclaimer to educate the user about misuse of the access token. They must acknowledge this disclaimer before they are able to copy the token. The disclaimer is a bit long, so it becomes scrollable if the max height for it is reached.
> 
>  output.mp4

I'm not going to approve this. I don't think this is good UI / UX and adding this token does nothing beneficial for our symposium.

frontend/src/components/navbar/navbar.module.css
```diff
@@ -30,7 +41,12 @@
   text-align: center;
 }
 
-.navLogout:hover,
+.navSettings:hover {
+  color: rgba(var(--button-rgb), 0.75);
+  cursor: pointer;
+}
+
+.navToken:hover,
```
AdrianJendo: Please also clean this up if you're not using it anymore

frontend/src/app/admin/settings/tokenConfirmation/page.tsx
```diff
@@ -0,0 +1,8 @@
+"use client";
+
+import TokenConfirmation from "@/components/settings/tokenConfirmation/tokenConfirmation";
+import React from "react";
+
+export default function SettingsModal() {
```
AdrianJendo: This is named incorrectly.

Don't `use client` and not needed to import React

frontend/src/app/admin/settings/page.tsx
```diff
@@ -0,0 +1,8 @@
+"use client";
+
+import Settings from "@/components/settings/settings";
+import React from "react";
+
+export default function SettingsModal() {
```
AdrianJendo: Don't `use client` and not needed to import React

backend/server.go
```diff
@@ -5,6 +5,7 @@ import (
 	"api/config"
 	"api/db"
 	"api/helpers"
+	authMiddleware "api/middleware"
```
AdrianJendo: Does this need to be aliased? Why wouldn't we call it `api/authMiddleware` if you feel the need to alias it

backend/middleware/auth.go
```diff
@@ -0,0 +1,156 @@
...
+		auth.ClearCookie(w)
+		if redirect {
+			http.Redirect(w, r, "/login", http.StatusSeeOther)
+		} else {
+			http.Error(w, "token expired", http.StatusUnauthorized)
+		}
+		return false
+	}
+	return true
+}
```
AdrianJendo: Is this whole file a move or did anything get changed?

backend/routes/webAuthn.go
```diff
@@ -13,5 +13,6 @@ func WebAuthnRoute() chi.Router {
 	r.Post("/authorize/request", controller.AuthorizeRequest)
 	r.Post("/authorize/response/{doorId}/{challenge}", controller.AuthorizeResponse)
 	r.Get("/authorize/{doorId}/{yubiKeyId}", controller.GetYubiKeyDoorAccess)
+	r.Get("/unlocked/{doorId}", controller.IsDoorOpen)
```
AdrianJendo, line 16: Did we make the necessary embedded changes to account for this?

backend/server.go
cameronkinsella: It's aliased because it conflicts with `github.com/go-chi/chi/v5/middleware`.  I don't mind changing it to `api/authMiddleware`.

backend/middleware/auth.go
cameronkinsella: The only change is it got wrapped with a `redirect` argument so that we can choose to redirect to the `/denied`, `/login`, etc. pages on certain conditions or just return a 401. We want that redirect behavior for frontend pages (e.g. visiting `/admin` if not admin should redirect to `/denied`), but we don't want that behavior for API requests (we just want to return a 401 for a POST request so they can see that there was an error).

backend/routes/webAuthn.go
cameronkinsella: Yes.

cameronkinsella:
Reverted settings modal and moved the access token dialogue to replace the logged in admin's self-delete button.
The button works the same as the video above.

image

frontend/src/app/admin/settingsModal/tokenConfirmation/page.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  const handleModalClose = () => {
+    router.back();
+  };
+
+  return (
+    <Modal
+      closeModal={handleModalClose}
+      title="Settings"
+      content={Confirmation()}
```
AdrianJendo: Is "Settings" the correct title? Also I don't think the brackets are necessary

frontend/src/components/navbar/navbar.module.css
```diff
@@ -14,10 +14,15 @@
   align-items: center;
 }
 
+.navToken {
```
AdrianJendo: Where is navToken used?

frontend/src/app/admin/settingsModal/tokenConfirmation/page.tsx
cameronkinsella: Changed the title to "Warning". The brackets on Confirmation are necessary because content expects an `Element` and `Confirmation` is `() => Element`.

frontend/src/components/navbar/navbar.module.css
cameronkinsella: Forgot to remove this. Updated.

AdrianJendo (approved):

PR #67, opened 2024-03-21
Redirect to requested page after login

cameronkinsella:
If you try to visit `/admin/foo` when your session is invalid, you will be redirected to `/admin/foo` after login.
If an admin initially visits `/` they will still be redirected to `/admin` after login.

### DIFF
--- a/backend/authMiddleware/auth.go
+++ b/backend/authMiddleware/auth.go
@@ -36,7 +36,7 @@ func VerifyAdmin(redirect bool) func(http.Handler) http.Handler {
 			if err != nil {
 				log.WithError(err).Warn("no token found")
 				if redirect {
-					http.Redirect(w, r, "/login", http.StatusSeeOther)
+					http.Redirect(w, r, "/login/?next="+r.URL.Path, http.StatusSeeOther)
 				} else {
 					http.Error(w, "no token found", http.StatusUnauthorized)
 				}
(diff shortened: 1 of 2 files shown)


backend/authMiddleware/auth.go
```diff
@@ -36,7 +36,8 @@ func VerifyAdmin(redirect bool) func(http.Handler) http.Handler {
 			if err != nil {
 				log.WithError(err).Warn("no token found")
 				if redirect {
-					http.Redirect(w, r, "/login", http.StatusSeeOther)
+					log.Info(r.URL.Path)
```
AdrianJendo: delete `log.Info` ?

frontend/src/app/login/page.tsx
```diff
@@ -45,7 +45,7 @@ const Login = () => {
             isAdmin: user.isAdmin,
           }),
         );
-        if (referrer) {
+        if (referrer && !(user.isAdmin && referrer === "/")) {
```
AdrianJendo: can't we just write `referrer && referrer !== "/"`

backend/authMiddleware/auth.go
cameronkinsella: oops 👀

AdrianJendo (approved):

PR #70, opened 2024-03-21
Add in ability to delete all fields in session data

ABarroso647:


backend/controller/webAuthn.go
```diff
@@ -353,6 +354,15 @@ func GetYubiKeyDoorAccess(w http.ResponseWriter, r *http.Request) {
 	helpers.JsonWriter(w, "Login Success, opening door")
 }
 
+func DeleteSessionData(w http.ResponseWriter, r *http.Request) {
+	res := db.DB.Session(&gorm.Session{AllowGlobalUpdate: true}).Delete(&models.SessionData{})
```
cameronkinsella: Why are we filtering by `AllowGlobalUpdate: true`?

ABarroso647: image
sql is as follows
we are not filtering 
image

cameronkinsella: ah, gotcha. Thanks.

ABarroso647: its just basically creating a session with delete all perms since they are not normally enables could also be done with 

db.DB.exec("Delete FROM session_data")

cameronkinsella (commented):
LGTM. 
@AdrianJendo can give it a quick check and approve.

backend/routes/webAuthn.go
```diff
@@ -13,6 +13,7 @@ func WebAuthnRoute() chi.Router {
 	r.Post("/authorize/request", controller.AuthorizeRequest)
 	r.Post("/authorize/response/{doorId}/{challenge}", controller.AuthorizeResponse)
 	r.Get("/authorize/{doorId}/{yubiKeyId}", controller.GetYubiKeyDoorAccess)
+	r.Delete("/sessionData", controller.DeleteSessionData)
```
AdrianJendo: @cameronkinsella should we put this in admin only routes?

cameronkinsella: Yeah. I'd prefer if we could add it under one of the base paths that are already private. All other endpoints in the /auth path are already public.

ABarroso647: not sure if we have a better fitting path tbh or a protected path that fits tbh. I could make like a /creds or smthn or throw it under user ig??? I do agree should be protected though

cameronkinsella: Yeah, it's hard to say. I would be fine with adding a new /creds path. You could set this endpoint as `/` for that path (i.e. `DELETE /creds`).
Thoughts? @AdrianJendo

AdrianJendo: That sounds good to me

backend/controller/creds.go
```diff
@@ -0,0 +1,18 @@
...
+	"net/http"
+)
+
+func DeleteSessionData(w http.ResponseWriter, r *http.Request) {
+	res := db.DB.Session(&gorm.Session{AllowGlobalUpdate: true}).Delete(&models.SessionData{})
+	if res.Error != nil {
+		http.Error(w, "Unable to delete session data", http.StatusInternalServerError)
+		return
+	}
+	helpers.JsonWriter(w, "Session data successfully deleted")
```
AdrianJendo: Can you return the # of records deleted, I think that would be helpful for debugging

backend/server.go
```diff
@@ -198,7 +198,7 @@ func api() chi.Router {
 		// Seek, verify and validate JWT tokens
 		r.Use(jwtauth.Verifier(auth.TokenAuth))
 		r.Use(authMiddleware.VerifyAdmin(false))
-
+		r.Mount("/creds", routes.CredsRoute())
```
AdrianJendo:
```suggestion
		r.Mount("/session", routes.SessionsRoute())
```

backend/routes/creds.go
```diff
@@ -0,0 +1,12 @@
+package routes
+
+import (
+	"api/controller"
+	"github.com/go-chi/chi/v5"
+)
+
+func CredsRoute() chi.Router {
+	r := chi.NewRouter()
+	r.Delete("/session", controller.DeleteSessionData)
```
AdrianJendo:
```suggestion
	r.Delete("/", controller.DeleteSessionData)
```

AdrianJendo (commented):
maybe its just me but `creds` isn't intuitive for me. Ill let @cameronkinsella decide what to do.

backend/controller/creds.go
cameronkinsella: +1

cameronkinsella:
A `/creds` route allows more flexibility for operations with the `credentials` table in the future, so my vote is for `/creds/session` or `/credentials/session`. I don't think we would do anything else with session, since most of the operations with `session_data` are part of the (public) webauthn protocol under `/auth`.

AdrianJendo:
> A `/creds` route allows more flexibility for operations with the `credentials` table in the future, so my vote is for `/creds/session` or `/credentials/session`. I don't think we would do anything else with session, since most of the operations with `session_data` are part of the (public) webauthn protocol under `/auth`.

Fair enough. I didn't think we'd do anything else with credentials so i thought maybe there's no point in overthinking it. I like `/credentials` more since its a real word but otherwise sgtm.

AdrianJendo (dismissed):

AdrianJendo (approved):

cameronkinsella (approved):

PR #71, opened 2024-03-22
Implement table select user settings

AdrianJendo:
Sorry it's so big, I just wanted to include all the last major frontend changes.

- Implement actions for selecting users
- Other misc quality of life changes / updates

https://github.com/lynx-locks/lynx-web/assets/55325093/0b492c51-77d2-4954-a276-ef3bf1c90f48

https://github.com/lynx-locks/lynx-web/assets/55325093/0e2818f0-9773-46b1-ac9a-fee1bdc124d0

### DIFF
--- a/backend/controller/user.go
+++ b/backend/controller/user.go
@@ -34,14 +34,14 @@ func GetAllUsers(w http.ResponseWriter, r *http.Request) {
 		}
 	} else {
 		// Split the comma-separated string into individual IDs
-		userIds := make([]uint64, 0)
+		userIds := make([]uint, 0)
 		for _, id := range strings.Split(userIdsParam, ",") {
-			intId, err := strconv.ParseUint(id, 10, 64)
+			intId, err := strconv.ParseUint(id, 10, 32)
 			if err != nil {
 				http.Error(w, "Invalid user ID", http.StatusBadRequest)
 				return
 			}
-			userIds = append(userIds, intId)
+			userIds = append(userIds, uint(intId))
 		}
 		db.DB.Find(&users, userIds)
 	}
@@ -130,6 +130,7 @@ func CreateUser(w http.ResponseWriter, r *http.Request) {
 	err := json.NewDecoder(r.Body).Decode(&user)
 	if err != nil {
 		http.Error(w, "Malformed request", http.StatusBadRequest)
+		return
 	}
 
 	ids := dbHelpers.GetAllIdsFromList(user.Roles)
@@ -163,6 +164,7 @@ func DeleteUser(w http.ResponseWriter, r *http.Request) {
 	err := json.NewDecoder(r.Body).Decode(&userIds)
 	if err != nil {
 		http.Error(w, "Malformed request", http.StatusBadRequest)
+		return
 	}
 
 	res := db.DB.Unscoped().Select(clause.Associations).Delete(&models.User{}, userIds.Users)
@@ -203,6 +205,7 @@ func DeleteUserCreds(w http.ResponseWriter, r *http.Request) {
 	err := json.NewDecoder(r.Body).Decode(&userIds)
 	if err != nil {
 		http.Error(w, "Malformed request", http.StatusBadRequest)
+		return
 	}
 
 	res := db.DB.Unscoped().Select(clause.Associations).Where("user_id IN ?", userIds.Users).Delete(&[]models.Credential{})
--- a/backend/controller/user.go
+++ b/backend/controller/user.go
@@ -34,9 +34,9 @@ func GetAllUsers(w http.ResponseWriter, r *http.Request) {
 		}
 	} else {
 		// Split the comma-separated string into individual IDs
-		userIds := make([]int, 0)
+		userIds := make([]uint64, 0)
 		for _, id := range strings.Split(userIdsParam, ",") {
-			intId, err := strconv.Atoi(id)
+			intId, err := strconv.ParseUint(id, 10, 64)
 			if err != nil {
 				http.Error(w, "Invalid user ID", http.StatusBadRequest)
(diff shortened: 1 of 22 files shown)


backend/controller/user.go
```diff
@@ -17,20 +17,31 @@ import (
...
+	userIdsParam := r.URL.Query().Get("users")
+	if userIdsParam == "" {
+		var err error
+		err, users = dbHelpers.GetAllTable(w, []models.User{})
+		if err != nil {
+			return
+		}
+	} else {
+		// Split the comma-separated string into individual IDs
+		userIds := strings.Split(userIdsParam, ",")
```
ABarroso647: does this give a proper error msg on string input?

ABarroso647: as in like incorrect input

backend/controller/user.go
```diff
@@ -138,14 +148,25 @@ func CreateUser(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	res := db.DB.Unscoped().Select(clause.Associations).Delete(&models.User{}, userIds.Users)
+	if res.Error != nil {
+		dbHelpers.DBErrorHandling(res.Error, w)
 		return
 	}
-	err = dbHelpers.DeleteObjandAssociationsByPk(w, models.User{Id: uId})
-	if err != nil {
+	if res.RowsAffected < 1 {
```
ABarroso647: should be erroring if we have less than inputted ids, no?

AdrianJendo: resolved

backend/controller/user.go
```diff
@@ -167,19 +188,20 @@ func GetUserCreds(w http.ResponseWriter, r *http.Request) {
...
 	w.Header().Set("Content-Type", "application/json")
-	err, uId := helpers.ParseInt(w, r, "userId")
+	userIds := struct {
+		Users []uint `json:"users"`
+	}{}
+
+	err := json.NewDecoder(r.Body).Decode(&userIds)
 	if err != nil {
-		return
+		http.Error(w, "Malformed request", http.StatusBadRequest)
```
ABarroso647, line 207: should still return here right?

backend/controller/user.go
AdrianJendo: If the input is incorrect it returns no users

backend/controller/user.go
```diff
@@ -167,20 +188,21 @@ func GetUserCreds(w http.ResponseWriter, r *http.Request) {
...
 	}
-	res := db.DB.Unscoped().Select(clause.Associations).Where(models.Credential{UserId: uId}).Delete(&[]models.Credential{})
+
+	res := db.DB.Unscoped().Where("user_id IN ?", userIds.Users).Find(&models.Credential{}).Delete(&[]models.Credential{})
```
ABarroso647: did we stest this one multiple creds? I had an account with 3 and it only deleted one

ABarroso647: verified it only deleted 1/2 the second time

backend/controller/user.go
ABarroso647: probably more ideal to actually parse it: https://stackoverflow.com/questions/24972950/go-convert-strings-in-array-to-integer 
this seems helpful

backend/controller/user.go
```diff
@@ -17,20 +17,39 @@ import (
...
+		var err error
+		err, users = dbHelpers.GetAllTable(w, []models.User{})
+		if err != nil {
+			return
+		}
+	} else {
+		// Split the comma-separated string into individual IDs
+		userIds := make([]int, 0)
+		for _, id := range strings.Split(userIdsParam, ",") {
+			intId, err := strconv.Atoi(id)
```
ABarroso647: sorry I didnt realize ud paste it exactly, still need to make it uint, I think u can just wrap unit(intvalue) not 100p

AdrianJendo: Atoi converts it to an int, why do we need it to be a uint?

backend/controller/user.go
```diff
@@ -167,20 +196,21 @@ func GetUserCreds(w http.ResponseWriter, r *http.Request) {
...
+	}{}
+
+	err := json.NewDecoder(r.Body).Decode(&userIds)
 	if err != nil {
-		return
+		http.Error(w, "Malformed request", http.StatusBadRequest)
 	}
-	res := db.DB.Unscoped().Select(clause.Associations).Where(models.Credential{UserId: uId}).Delete(&[]models.Credential{})
+
+	res := db.DB.Unscoped().Select(clause.Associations).Where("user_id IN ?", userIds.Users).Delete(&[]models.Credential{})
```
ABarroso647, line 211: can u paste new sql output? this looks like it'll perform better tho perhaps (well I'm assuming u tested it)

backend/controller/user.go
ABarroso647: cuz all our ids are uints?

ABarroso647: if u look at the helper function that normally does this its always converting to uint

backend/controller/user.go
AdrianJendo: It was DELETE * FROM credentials WHRE user_id in ()

backend/controller/user.go
ABarroso647: unresolved, it does not return here

backend/controller/user.go
ABarroso647: perfect, ty

backend/controller/user.go
ABarroso647: can you also add it in deleteUser, that is probably my bad tho

cameronkinsella (approved):

PR #73, opened 2024-03-22
Add button to clear session data

AdrianJendo:
- Figured i would put it in the admin's settings modal to keep it simple

Screenshot 2024-03-22 at 1 07 56 AM

### DIFF
--- a/frontend/src/app/admin/settingsModal/page.tsx
+++ b/frontend/src/app/admin/settingsModal/page.tsx
@@ -109,6 +109,13 @@ export default function SettingsModal() {
     }
   };
 
+  const handleRefreshSessionData = async () => {
+    if (confirm("Are you sure you want to refresh the session data?")) {
+      await axios.delete(`/credentials/session`);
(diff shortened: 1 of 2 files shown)


frontend/src/app/admin/settingsModal/page.tsx
```diff
@@ -181,6 +186,14 @@ export default function SettingsModal() {
...
                 </button>
               )}
+              {user?.id === settingsUser.id && (
+                <button
+                  className={styles.deleteButton}
+                  onClick={handleRefreshSessionData}
+                >
+                  Refresh Session Data
+                </button>
+              )}
```
cameronkinsella, line 196: Can't we put this button next to the Access Token button to avoid the extra `user?.id === settingsUser.id`?

AdrianJendo: Because then the styling is more effort because i need another `<div>`

cameronkinsella (approved):

PR #75, opened 2024-03-23
Remove auth flow Search Params

AdrianJendo:
Since we're using static doorIds for the authorization flow, I think it's useful to try to hide them from the client as much as possible. This change loads the doorId into state and removes them from the query parameters.

### DIFF
--- a/frontend/src/components/successMessage/successMessage.module.css
+++ b/frontend/src/components/successMessage/successMessage.module.css
@@ -1,11 +1,5 @@
-.successMessageContainer {
-  display: flex;
-  height: 100vh;
-}
-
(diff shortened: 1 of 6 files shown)


frontend/src/components/successMessage/successMessage.module.css
```diff
@@ -1,11 +1,5 @@
-.successMessageContainer {
-  display: flex;
-  height: 100vh;
-}
-
```
AdrianJendo, line 5: This styling looked better on desktop, but it looks worse on mobile so I think we're better off reverting it.

cameronkinsella (approved):

